Issue #1, opened 2019-05-30
Minor additions and modifications to developers documentation not worth a separate ticket


abuts:
Here we commit small changes and modifications which do not worth a separate ticket

PR #3, opened 2020-02-18
Split optmodel

g5t:
Thanks to the comments from @mducle I have made rather significant changes to the design.

Most of the `OptModel` properties have been consolidated into three `OptFunctionsArray` objects, each of which contains a list of `OptFunction` objects, any inter-function parameter binding information, the function-to-dataset `applies` information, and how multiple functions that apply to a single dataset should be combined.

Additional modifications have been made to address issues raised in the [numbered list](https://github.com/pace-neutrons/Pace-Project-Plan/issues/1#issuecomment-587125360) namely

1. Clarification that `OptFunction` objects represent a 'fixed' function which *should not* be modified, though we do not need to enforce real immutability since a determined user will probably always be able to redefine an underlying function by, e.g., editing an m file.
2. Clarification that the `OptFunction.parameter_functions` specify the bindings between the function parameters.
3. Specified that we anticipate only `numeric` or `sqw`-like `input_type` for `OptFunction` functions, and that we might be able to do-away with this object property in C++ and Python implementations.
4. Clarified that only a predefined set of optimisation back-ends will be supported unless a general interface becomes apparent during implementation. This may require that users submit pull requests adding the interface to their preferred optimisers.
5. Added the possibility that OptFunction parameters might be named by the user, and that such names could be used in specifying parameter bindings and limits. The syntax for this is not yet fixed.


optimisation/design/Model_Optimisation_Design.md
```diff
@@ -253,87 +266,22 @@ length vector,
...
 
-## Parameter binding
-We should use functions to enforce binding of parameters.
+| required number | pass to function |
+|-----------------|------------------|
+| 0               | `void`           |
+| 1               | `E`              |
+| 2               | `norm(Q),E`      |
+| 3               | `h,k,l`          |
+| 4               | `h,k,l,E`        |
```
mducle, line 302: It seems to me that implicit in this design is the assumption that the user defined function will give only a **single** vector output of the same size as its input coordinate vectors.

The use-case I'm thinking of is when the user defined model function takes three input vectors - specifically the case of `Euphonic` and `SpinW`. it's true that they both require the `h`, `k`, and `l` vectors which `OptFunction` will supply to them, but in both cases what is returned is actually not a single vector `s` of the same size as the input vectors, but *N_b* (number-of-branches) sets of pairs of vectors. One of pair would be the dispersion relation `w` (energy of modes as a function of `h,k,l`) and the other is the scattering intensity `s`.

Currently in Horace this output (for `SpinW` at least; I don't know how `Euphonic` is treated) is converted to a single `s` vector for a given `h`, `k`, `l` *and* `e` input vectors by the `disp2sqw` function which applies an energy broadening to each mode and sums the resulting "convolutions" (actually more a smoothing filter). The default broadening function in `disp2sqw` is a Gaussian but it also accepts a user provided function. 

The question is how we provide this functionality in `OptFunction` which seems to me to be more a translation / update of `multifit_sqw` (which always expects its user defined model functions to output a single vector with inputs `h`, `k`, `l`, `e`). Should we have (yet another) helper class to do the equivalent of `disp2sqw`? Or should it be done in the `OptModel` class?

optimisation/design/Model_Optimisation_Design.md
```diff
@@ -253,87 +266,22 @@ length vector,
...
```
mducle, line 303: The other issue is that if you're fitting a `dnd` object rather than an `sqw` object - that is, a `dnd` object is passed to `OptFunction` as input instead of input vector arrays - the it might be impossible to extract what `h`, `k`, `l` are (`e` will generally always be discernable / defined). In the current implementation this is handled by `multifit` (not `multifit_sqw`) by ignoring what the `h,k,l` might be and just passing the `dnd` object's coordinates directly to a fit function (generally some kind of peak function or polynomial which don't really care about `h,k,l`.

So the question here is whether or how this case should be handled - e.g. what happens if the user defined function accepts only one input coordinate vector but we are passed a `d2d` (2D) object with two coordinate vectors. Should we raise an error, or just pass the first vector (implicitly averaging over the second dimension)?

optimisation/design/Model_Optimisation_Design.md
```diff
@@ -433,23 +364,158 @@ the bounding functions specified might be of the form:
...
+## Optimisation Functions Array Object
+It is worthwhile to allow for a user defined ![S(\mathbb{Q};\mathbf{p})] to be constructed as the sum of a number of user defined functions held in `OptFunction`s.
+And when fitting multiple datasets it will be worthwhile to have a list of `OptFunction` objects each of which is associated with one or more of the datasets.
+This will allow, e.g., multiple datasets to be fit to related functions where some number of their parameters are bound by user defined relationships.
+A new class is necessary to hold multiple `OptFunction` objects and handle their interactions and application to multiple datasets, `OptFunctionsArray` 
+
+The minimum set of properties of an `OptFunctionsArray` object are
+
+- a list of one or more `OptFunction` objects
+- function to dataset applicability information
```
mducle: missing a verb here... ("function to **[hold / encapsulate / denote]** dataset applicability information")

optimisation/design/Model_Optimisation_Design.md
```diff
@@ -433,23 +364,158 @@ the bounding functions specified might be of the form:
...
+[\mathbf{p}_i]: https://latex.codecogs.com/svg.latex?%5Cmathbf%7Bp%7D_i
+[\mathbf{p}_j]: https://latex.codecogs.com/svg.latex?%5Cmathbf%7Bp%7D_j
+[\mathbf{p}_k]: https://latex.codecogs.com/svg.latex?%5Cmathbf%7Bp%7D_k
+
+such that they collectively describe a real valued function
+
+![\rho(a,\ldots;\mathbf{p}) = M(a,\ldots;\mathbf{m}) \times F(a,\ldots;\mathbf{f}) + B(a,\ldots;\mathbf{b})]
+
+[\rho(a,\ldots;\mathbf{p}) = M(a,\ldots;\mathbf{m}) \times F(a,\ldots;\mathbf{f}) + B(a,\ldots;\mathbf{b})]: https://latex.codecogs.com/svg.latex?%5Crho%28a%2C%5Cldots%3B%5Cmathbf%7Bp%7D%29%20%3D%20M%28a%2C%5Cldots%3B%5Cmathbf%7Bm%7D%29%20%5Ctimes%20F%28a%2C%5Cldots%3B%5Cmathbf%7Bf%7D%29%20&plus;%20B%28a%2C%5Cldots%3B%5Cmathbf%7Bb%7D%29
+
```
mducle, line 504: What about the case (just saying!) that I want to a different multiplier function for each of (or perhaps a subset of) foreground functions? E.g. where I want:

\rho(a,...; p) = \sum_j M_j(a,...; m) F_j(a,...; f) + \sum_k B_k (a,...; b)

instead of:

\rho(a,...; p) = \prod_i M_i(a,...; m) \times \sum_j F_j(a,...; f) + \sum_k B_k (a,...; b)

?

Should we cater for this use case, or should we explicitly rule it out? In principle if the foreground functions are physical models of scattering (e.g. magnons and phonons) they *should* be outputting absolute units so should not need a scaling factor between them, only a single scaling factor for the data (which may or may not have been normalised to a standard measurement).

optimisation/design/Model_Optimisation_Design.md
```diff
@@ -473,14 +539,13 @@ bg = [OptFunction(@linear_in_h_bkg, hklT{:}, 'parameters', [1,0]), ...
...
-opt.set_foreground_functions(fg);
-opt.set_background_functions(bg);
+% Create the OptModel object
 % the first background function should be treated independently for each sqw
 % and all other functions apply to all sqw objects
-opt.set_background_applies({-1,0,0});
+opt = OptModel([sqw_object_1, sqw_object_2], ...
+               'foreground', [sw_func, ph_func],...
+               'background', OptFunctionsArray(bg, 'applies', {-1,0,0}), ...
+               'multiplier', multiplier);
```
mducle, line 588: I guess here `OptModel` implicitly constructs a 2-element `OptFunctionsArray` for the input `[sw_func, ph_func]` and single-element `OptFunctionsArray` for the input `multiplier`?

optimisation/design/Model_Optimisation_Design.md
```diff
@@ -528,6 +594,7 @@ If model optimisation is to be supported when running compiled MATLAB/Horace fro
 - Can compiled MATLAB call pure Python functions?
 	+ How should such functions be provided to MATLAB? As `module.function` strings, memory addresses, etc.?
 - Does compiled MATLAB have access to builtin MATLAB functions like `feval` and `str2func`?
+	+ Either way it can not run arbitrary m-files! No MATLAB user functions can be called using the Python interface.
```
mducle, line 640: In principle yes... 

With code compiled using R2017b, all Matlab built-in functions and all included library functions are callable using:

```python
>>> import compiled_matlab_library
>>> m = compiled_matlab_library.initialize()
>>> m.eval('eig(rand(3))')
matlab.double([[-0.18794383321803343],[1.7526703425428873],[0.8398632587280312]])
```

However, newer versions of Matlab restrict access to only specifically denoted m-files. In principle all Matlab functions, including mex and built-ins, are still accessible within the Matlab process even for the compiled library but now we have to write a single-m-file wrapper class to allow arbitrary Python access to the library functions.

On the minus side this will be a pain but we would probably want a Python-side wrapper anyway to make the syntax nicer for users, so it's not *that* much pain... On the plus side it shouldn't matter what class types (value or handle) Horace/Herbert/PACE Matlab classes are since we will need to write a (handle) class wrapper for them anyway...

optimisation/design/Model_Optimisation_Design.md
```diff
@@ -528,6 +601,7 @@ If model optimisation is to be supported when running compiled MATLAB/Horace fro
...
```
mducle, line 641: One very important "gotcha" with the compiled Matlab is:

```python
>>> m.eval('py.dict()')
Warning: Ability to call Python libraries from MATLAB is not available with MATLAB Compiler SDK for Python.
```

(the same code works [e.g. calling python from Matlab from within python] if you have a Matlab license, but not with the compiled version!). 

This error also appears if a newer version of Matlab is used to compile the code - so to have Python user model functions we need to use some sort of C++ or inter-process communications work-around which will be a major pain.

mducle (changes requested):
Just a "few" comments but otherwise looks good.

Could you add the compiled Matlab information? (The other comments don't really warrant any changes to design documents... maybe). I'm planning to write a more detailed document for compiled-matlab / python interfacing but it would be good to have it here for discussion first.

optimisation/design/Model_Optimisation_Design.md
mducle: PS. I think "eluded" should be "alluded" (wherever it is... I missed the push which added that sentence).

optimisation/design/Model_Optimisation_Design.md
g5t: My idea was that functions which take the components of `Q` and return mode information at each `Q` would be handled exclusively *with* resolution convolution, possibly like the 'single mode approximation' in [ResLib](https://neutron.ethz.ch/research/resources/reslib.html).

With the proposed addition of a `convolute` flag (or structure) to `OptFunction`  it seems there could be a clear operation to perform if no convolution is to be performed *and* the function takes the components of `Q` and returns mode information -- namely converting from modes to a continuous function evaluated at each (Q,E).

optimisation/design/Model_Optimisation_Design.md
g5t:
``` proposal
- function-to-dataset mapping/applicability information
```

g5t:
```suggestion
- function-to-dataset mapping/applicability information
```

optimisation/design/Model_Optimisation_Design.md
g5t: I think it is O.K. to require that all foreground functions that apply to a single dataset should return intensity in the same units. If a user wanted to combine two models with different units, say `mb/Sr/meV/f.u.` vs. `mb/Sr/meV/atom` they would need to wrap one model in a simple function that applies the appropriate conversion factor.

If you disagree with my opinion then we can add the possibility of per-foreground-scale-factors in one of two ways

- a more general construction of &rho;(a,...;p) at the expense of extra bookkeeping
- providing *all* functions with an overall scale-factor as part of `OptFunction` that could be a single value (defaulting to 1) or another `OptFunction`

optimisation/design/Model_Optimisation_Design.md
g5t: Implicit in this notation is that `OptModel` automatically constructs `OptFunctionsArray` objects from scalar or array `OptFunction` objects. The automatically constructed `OptFunctionsArray` objects would have default parameters (to be determined) so any non-default settings might be easiest accomplished by constructing the `OptFunctionsArray` object by hand, as in the case of the `'background'` functions.

optimisation/design/Model_Optimisation_Design.md
g5t: Do we want to continue to support fitting models without direct physical interpretations?
Yes, you can fit a series of Gaussian peaks and use those to, e.g., extract a spinwave velocity **or** you could fit the peaks simultaneously as part of a model that *includes* the spinwave velocity as a parameter.

If we do want to continue to support fitting data based on its binning axes then we will need to decide on a more-complex `input_type` parameter for `OptFunction` (maybe `'coords'`, `'sqw`', and `'axes'` instead of, e.g., `'numeric'` and `'object'`)

optimisation/design/Model_Optimisation_Design.md
mducle: I agree - just wanted to raise the point

optimisation/design/Model_Optimisation_Design.md
```diff
@@ -632,6 +639,7 @@ The advantages and disadvantages of each potential language need to be assessed
 | C++      | compiled &rarr; functionality constrained &rarr; fully testable | increased distribution complexity                    |
 |          | Python interface via `pybind11` simple                    | MATLAB interface via `mex` less simple                     |
 
+The interpreter issues eluded to above is an uncertainty regarding what happens if an interactive Python interpreter calls a function in the compiled MATLAB/Horace which in turn calls functions in a Python module?
```
g5t:
```suggestion
The interpreter issues alluded to above is an uncertainty regarding what happens if an interactive Python interpreter calls a function in the compiled MATLAB/Horace which in turn calls functions in a Python module?
```

abuts (approved):
Its highly involving document, providing overview of the desired fitting framework. I have no objections in general and the document is not suitable for all small details, which will actually make the difference. 
Couple of notes:
1)	Current interface is clear and suitable for number of basic operations (except binds settings – nobody will miss this) so it should remain and be transferrable to new interface under the hood. 
2)	The fitter (OptModel) replacement involving different optimization methods (except Levenberg-Marquardt) and the related issue of the fitter interface are not covered by the document.
3)	Fitting with and without resolution convolution is assumed to be changed by a switch. By design, these two are pretty different and complex classes with common interface. Switch can be implemented on a user interface level, though two OptModel-type classes doing these kind of fittings like it is done now may be beneficial. 
4)	… these are all minor remarks and I can find  number of others, but I do not think, will add a lot at this stage.

mducle (approved):
Looks good to me now!

optimisation/design/Model_Optimisation_Design.md
mducle: It was decided in the sprint meeting that we *should* continue to support fitting models without physical coordinate axes (e.g. allow fitting based on binning axes).

optimisation/design/Model_Optimisation_Design.md
mducle: I'll try to answer these issues another design document and will make PR to update the text here when the Matlab-Python interface is better understood.

optimisation/design/Model_Optimisation_Design.md
g5t: Thanks, I meant to add a similar comment about the groups' decision but forgot. 😆
The newest version of the design document explicitly includes this possibility.

PR #4, opened 2020-02-26
Add initial Euphonic v0.3 design

rebeccafair:
Addresses [this issue](https://github.com/pace-neutrons/Pace-Project-Plan/issues/49)


euphonic/design/03_design_document_v0.3.md
```diff
@@ -0,0 +1,385 @@
...
+spectra for (usually powdered) molecular crystals, for indirect geometry
+instruments. Euphonic could improve the simulated spectra in Abins by allowing
+calculations to be made from data at any q, and provide a good platform for
+methods development. Abins will still implement its own plotting and instrument
+specific broadening and handle interaction with Mantid.
+
+In the literature these molecular crystal vibrations are usually separated into
+'internal' molecular vibrations (which are assumed to be flat) and 'external'
+vibrations (translations or rotations of the entire molecule). Currently Abins
+assumes all vibrations are flat (no dispersion) and employs semi-analytic power
```
g5t:
```suggestion
assumes all vibrations are flat (no dispersion) and employs semi-analytic powder
```

euphonic/design/04_algorithms.md
```diff
@@ -0,0 +1,89 @@
...
+&alpha;, M<sub>&kappa;</sub> is the mass of atom &kappa;, r<sub>a</sub> is the
+vector to the origin of cell a in the supercell,
+&epsilon;<sub>q&nu;&kappa;&alpha;</sub> are the eigevectors, and
+&omega;<sub>q&nu;</sub><sup>2</sup> are the frequencies squared.
+
+If the material is polar (has non-zero Born charges) there is a long-range
+component of the force constants matrix due to long-ranged dipole-dipole
+interactions within the material. This can be accounted for by applying a
+(computationally expensive) correction to the dynamical matrix. This calculation
+is relatively complex and detailed in <sup>[2]</sup>, and a shorter summary can
```
nickbattam-tessella: If you're just giving references by number they probably don't need to be superscripted (`[2]`). 
If you're using superscripts I'd detail the reference too e.g. `Gonze & Lee<sup>[2]</sup>` or just reference `Gonze & Lee (1997)`

euphonic/design/04_algorithms.md
```diff
@@ -0,0 +1,89 @@
...
+and eigenvectors at q. The eigenvalue equation is:
+
+<p><img align="center" src="http://chart.apis.google.com/chart?cht=tx&chl=D_{\alpha{\alpha}'}^{\kappa{\kappa}'}(q) \epsilon_{q\nu\kappa\alpha}=\omega_{q\nu}^{2} \epsilon_{q\nu\kappa\alpha}"/> (1.3)</p>
+
+Where &nu; runs over phonon modes, &kappa; runs over atoms, &alpha; runs over
+the Cartesian directions, a runs over unit cells in the supercell,
+u<sub>&kappa;&alpha;</sub> is the displacement of atom &kappa; in direction
+&alpha;, M<sub>&kappa;</sub> is the mass of atom &kappa;, r<sub>a</sub> is the
+vector to the origin of cell a in the supercell,
+&epsilon;<sub>q&nu;&kappa;&alpha;</sub> are the eigevectors, and
```
nickbattam-tessella, line 26: your epsilons are inconsistent; you have `\epsilon` in the equation and `\varepsilon` in the text (using \LaTeX notation)

euphonic/design/03_design_document_v0.3.md
```diff
@@ -0,0 +1,385 @@
...
+have to lie elsewhere in the code, and dictionary access (`obj['a']`) would
+clash syntactically with attribute access (`obj.a`) elsewhere.
+
+**Numpy Array** - Very simple, but wouldn't be able to bundle associated data
+together and would require the user to keep track of this. For serialisation,
+plotting, further calculations the user would have to pass all pieces of
+required data as arguments which is messy and likely to cause errors.
+
+## Units
+Units in Euphonic will be handled by the [`Pint`](https://pint.readthedocs.io)
```
nickbattam-tessella, line 465: it's unclear from this what (if any) units the user has to pass data into Euphonic in.
I remember from the review meeting that it was required that data be passed in using Pint wrappers in the same Pint-namespace so whatever conversions are needed can be made.

euphonic/design/04_algorithms.md
```diff
@@ -0,0 +1,89 @@
...
+q-direction of approach to gamma and results in LO-TO splitting. This is also
+described in <sup>[2]</sup> and <sup>[3]</sup>.
+
+
+## 2. Calculation of the crystal structure factor
+
+The crystal structure factor describes how phonons in the material scatter
+incident neutrons. It is calculated for each Q (note: capital Q is neutron
+momentum transfer, small q is phonon wavevector) and phonon mode &nu; by
+<sup>[4]</sup>:
```
nickbattam-tessella: I'm not sure your references need to be superscripted if you're just referencing by number. If you used `Dove (2003)<sup>[4]</sup>` would make sense

euphonic/design/04_algorithms.md
nickbattam-tessella, line 26: your epsilons are inconsistent between the equations (\epsilon) and text (\varepsilon) (in LaTeX notation). You can change the text one to a non-curly epsilon by switching from `&epsilon;` to `&#x3F5;` or the equation to curly `\epsilon` to `\varepsilon`

euphonic/design/03_design_document_v0.3.md
```diff
@@ -0,0 +1,385 @@
...
+|x bin units|string|
+|y bin edges|(n_ybins + 1) array of floats|
+|y bin units|string|
+|x tick labels|list of strings|
+|x tick locations|array of ints|
+|plot data|(n_xbins, n_ybins) array of floats|
+
+|Method|Input required|Output|
+|------|--------------|------|
+|Plot|x,y broadening|Matplotlib Figure|
```
nickbattam-tessella: The `plot` method in here (and the `1D Spectrum`) should be in the 1D/2D plotter and not in the spectrum (data) objects. Having it makes it hard to implement the separation you talk about in the Plotting section below
```
Plotting functions will be contained in their own separate module, so that libraries such as `Matplotlib` will only be imported if a user wants to plot, and not required for general data analysis.
```

ajjackson: I do think there should be a `.plot()` method on the data objects; it's too cumbersome to go and find the appropriate plotting function and call it on an object, especially given that these will be relatively simple plots. The method could dispatch to (otherwise-useless) code living in another module, but that's a housekeeping detail really; the point is that either way you call `plot()`, _then_ it attempts to import matplotlib and raises an ImportError if you don't have it. You can safely instantiate the object, examine it and call `serialise()` without matplotlib.

nickbattam-tessella: `1dplotter.defaultplot( spectrum ) === spectrum.plot()`

As soon as plots become part of the nD-spectrum objects they become more complex; they are data objects that CAN BE PLOTTED by whatever library you want to use

euphonic/design/03_design_document_v0.3.md
```diff
@@ -0,0 +1,385 @@
...
+object definition.
+
+|Attribute|Type|
+|---------|----|
+|x bin edges|(n_xbins + 1) array of floats|
+|x bin units|string|
+|y bin edges|(n_ybins + 1) array of floats|
+|y bin units|string|
+|x tick labels|list of strings|
+|x tick locations|array of ints|
```
ajjackson: If the x tick locations/labels are always used together to set special points, maybe they should be a single attribute: a list of 2-tuples `[(loc1, label1), (loc2, label2), ...]`. We should _not_ use a dict of `{label: loc}` because repeated labels are quite common. `{loc: label}` would also be unsafe as multiple labels might be set at the same x-position to represent jumps from one special point to another. And then if the attribute is set to `None` we know to fall back on default x-labelling based on q values; it's not obvious what should happen if *x_tick_locations=[1, 2 3]* and *x_tick_labels=None*.

An edge case we hit in Sumo is that you can't even rely on a label always referring to the same vector in **q**-space; between symmetry and periodicity it might make sense for a path to cover more than one equivalent special point. The band structure format used by Pymatgen, which stores actual q-points as well as a dict mapping special points to labels, blows up in this scenario; we had to hack around it by creating new symmetry labels that look the same when printed.

euphonic/design/03_design_document_v0.3.md
ajjackson: What scenario do you have in mind where that added method creates a maintenance problem? It has no impact on serialisation or at interfaces between languages, because only the stored data needs to be exchanged.

`import 1dplotter; 1dplotter.defaultplot(spectrum)` is an objectively worse user experience from an interactive shell or notebook than `spectrum.plot()`. If there's a maintenance benefit then maybe the trade-off is worth it, but I'd like to understand what it is so I can avoid the same mistake in future.

nickbattam-tessella: It's not a "mistake" and I imagine counter arguments could be made for its inclusion.

I am thinking primarily of extra maintenance effort and additional cognitive load for both maintainers and users arising from the complexity of having plot functionality on the spectrum class AND the separate plotter object.

1) Plotting is be done in two places in the codebase via to distinct APIs. 
```
spectrum.plot();
```
and 
```
import plotter;
plotter.plot(spectrum);
```
This raises questions for a user which one should I be using? Is there a difference? 
For developers, if the 'default' plot is available in both there is a maintenance overhead of keeping them inline. If I want to change some style on all Euphonic plots I have to do it in the spectrum plot as well as the plotting classes.

2) The spectrum class now does two distinct things:  calculates/stores data about spectra (backend functionality) AND plots spectra, popping open a UI element to do it.
In contract the plotter class does one thing: provide (multiple) standard plots of spectrum data

3) A user  is likely have their own preferred plots and plotting libraries which they will want use for custom plots for publication or to look at data not covered by the standard set. These would need to work with an appropriately dimensioned spectrum object. By doing all the Euphonic plots via the plotter class(es) we have ensure that all data is available through the API in a usable form.

4) A lazy-import of matplotlib removes the issue of running on a machine without that library installed - if that plot routine is never called, but calling `plot` on spectrum would then fail; if you're always plotting through the plotter class that dependency becomes more explicit and your script would fail early.

The issue of having plotting routines "outside" the main analytic code is something that Greg has raised with Horace which bundles a load of plots he doesn't want in.

euphonic/design/03_design_document_v0.3.md
nickbattam-tessella: Are x-tick locations int values to point number or a position? 
Python `zip` would neatly join the two lists of labels+locations for use in a loop or something if two lists are easier to manage than a list of tuples.

euphonic/design/03_design_document_v0.3.md
ajjackson: Thanks for explaining, that makes sense. Redundancy is bad, as are bloated God-objects. My take:
- If the plotter for a 1D/2D spectrum is fairly simple then it should live on the class and not be redundantly callable from somewhere else.
- If we have a whole assortment of plotting tools for this object then it is reasonable for them to live elsewhere.

I was under the impression that the former is more accurate, as plotting is more of a convenience/diagnostic tool and not a major role of Euphonic.

> If I want to change some style on all Euphonic plots I have to do it in the spectrum plot as well as the plotting classes.

I agree with the principle here, but for this specific example it seems good to mention:
If we are doing any plot styling, we should use Matplotlib's cascading style system. Not only does it make this problem go away, it also cuts down on the need to repeat ourselves between "unrelated" plotters that can share a base stylesheet.

> 2. The spectrum class now does two distinct things: calculates/stores data about spectra (backend functionality) AND plots spectra, popping open a UI element to do it.

The spectrum classes don't do any calculating, only storage. And an object-oriented plotter naturally stores plot data, so why not have them be the same object?

Point 3 is an interesting one; basically, having an extra step where we use our own API is a good thing because it keeps us honest? We should be able to rely on testing to make sure the API behaves as advertised.

Point 4: It does suck to have a script die to surprise dependencies, so this seems a good design principle. This particular case is an *especially* unsurprising dependency; I don't think users can get mad at us because their script calling a method named `plot()` blows up without matplotlib. But there remains a theoretical risk that we will mess up by implicitly calling plot() as part of something else. If we are saying matplotlib is optional then we should make sure to have some CI that runs on a matplotlib-less container.

euphonic/design/03_design_document_v0.3.md
ajjackson: `zip` won't complain if the lists are different lengths, it will quietly skip the excess entries. (This is very handy if one of the iterators is infinite!)

euphonic/design/03_design_document_v0.3.md
rebeccafair: Thanks both for your feedback and discussion.

As the plotting in Euphonic should be kept simple, I was leaning towards just having a plot method on the object, however there is one particular use case which will require a separate plotter module: plotting multiple 1D spectra on the same axes. This can't be done with a plot method, and we've agreed being able to plot in multiple places is a bad idea. So we'll go with the plotter module and I'll update the doc to make that clearer.

euphonic/design/03_design_document_v0.3.md
rebeccafair: x-tick locations `int` values are to point to a bin number.

I agree that a list of tuples would be better than 2 lists, I'll update this

euphonic/design/03_design_document_v0.3.md
rebeccafair: I've been playing with dataclasses/classes and pint units and I think if a user is instantiating an object directly, they should wrap any data with units as a pint `Quantity` object (which would be made clear in the docstring). What is less clear is what we should do for intput/output, from files or other data structures such as dictionaries. One option is to output the data as a numpy array with a string specifying the units e.g. `cell_vectors=np.ndarray([...])` and `cell_vectors_unit='angstrom'`, however there are also options for converting `Quantity` objects to a tuple of magnitude and unit for file output, which I'm also looking into, but this could potentially be inefficient.

I've done a quick prototype of the `Structure` object mentioned above which shows how this could work, which I'll upload and I'll generally update the doc to make this clearer. On another note my prototype shows that classes might actually be better than dataclasses for dealing with this unit stuff...

euphonic/design/03_design_document_v0.3.md
ajjackson: Is that the index of a bin _edge_ or a bin _midpoint_?

rebeccafair: Bin midpoint, as for the use case of labelling high-symmetry points there would be 1 q-point per bin and you would want to label that particular q-point.

However, if the axis were an energy axis, it's probably more likely you would want to label bin edges, but I imagine this use case would be covered by Matplotlib's default tick labels? I can't really think of a better way to handle this...

euphonic/design/03_design_document_v0.3.md
```diff
@@ -0,0 +1,531 @@
...
+cheaply get frequencies and eigenvectors on a finer grid via linear
+interpolation. The structure factor can then be calculated on this finer grid
+(calculating the structure factor is much cheaper than calculating the
+frequencies/eigenvectors). The hierarchy of Horace/Brille/Euphonic is not yet
+decided so Euphonic must be flexible. In addition, in the future there is scope
+to calculate the structure factor for powder samples, this also needs to be
+considered. From this the requirements are:
+
+- Euphonic must be able to take a list of q-points as input, and return the
+crystal structure factor, and phonon frequencies for each q-point
```
ajjackson, line 31: It's not clear from this description how Euphonic should get the necessary information about the lattice dynamics. Does Horace 
- also specify the path to a phonon data file,
- pass a previously instantiated object containing lattice parameters and force constants or
- require Euphonic to hold an internal state that has previously been set to the system of interest?

euphonic/design/03_design_document_v0.3.md
```diff
@@ -0,0 +1,531 @@
...
+specific broadening and handle interaction with Mantid.
+
+In the literature these molecular crystal vibrations are usually separated into
+'internal' molecular vibrations (which are assumed to be flat) and 'external'
+vibrations (translations or rotations of the entire molecule). Currently Abins
+assumes all vibrations are flat (no dispersion) and employs semi-analytic powder
+averaging ([1]) up to 4th order multiphonon interactions. This could
+potentially be improved by integrating the eigenvectors over the Q accessible by
+the instrument, rather than assuming a flat bandstructure, but the exact method
+for doing this is not yet clear. If may also be beneficial to perform a
```
ajjackson: Would be worth mentioning that the current method models _incoherent_ scattering and a numerical powder average would provide access to coherent effects.

euphonic/design/03_design_document_v0.3.md
```diff
@@ -0,0 +1,531 @@
...
+numerical powder average rather than semi-analytical, but this may make
+calculating higher order interactions prohibitively expensive, and needs to be
+investigated. These calculations should produce a 1D spectrum in energy,
+resolved per atom and quantum order. For Euphonic to be useful to Abins, the
+requirements are:
+
+- Euphonic must be able to produce a 1D energy-binned incoherent spectrum
+resolved per atom and quantum order
+- Euphonic must run with versions of Python that are shipped with Mantid (as of
```
ajjackson, line 72: Additional requirement: This should not require access to Hessian/Force constants as Abins currently supports input file formats that only provide frequency and eigenvector information. Those file parsers can stay in Mantid, but the 1D incoherent calculator should be able to take a series of energy/q/vector as input.

euphonic/design/03_design_document_v0.3.md
rebeccafair: This part is mainly describing what the required inputs/outputs are for a Horace simulation function. You must pass to Horace a function which takes qh, qk, ql at a minimum (plus whatever parameters you like) and returns the structure factor, and mode frequencies at those q-points. The extra parameters to that function and what happens inside it are sort of implementation details and don't matter to Horace.

In reality we'd probably get a Horace user to input a model type (e.g. 'CASTEP' or 'Phonopy') then allow whatever keyword arguments apply to the from_castep, from_phonopy methods, where Euphonic would then read and interpolate the force constants. But that could be changed depending on if Brille is an intermediate, we end up needing to write to file etc... As long as Euphonic can take q-points and return structure factors and mode frequencies that's all Horace requires.

euphonic/design/03_design_document_v0.3.md
```diff
@@ -0,0 +1,531 @@
...
+# Current Solution and Shortcomings
+
+See [v0.2.2 overview](01_design_overview_v0.2.2.md)
+
+# Proposed Solution
+
+## Euphonic Data Flow
+
+<img align="center" src="images/03_data_flow.png">
+
```
ajjackson, line 130: This data flow suggests that Abins does not depend on Brille, because Euphonic never calls Brille. That is helpful from a packaging perspective, but removes the option to use Brille as part of Euphonic's powder-averaging workflow.

Do we have a firm decision to allow/disallow calling Brille from Euphonic?

euphonic/design/03_design_document_v0.3.md
ajjackson: So the workflow is actually:
- pass phonon data to Euphonic
- get back a function object `f(q) -> SparseScatteringData`
- call `f(my_qpts)` from Horace?

Would it be more accurate to say "For a given set of lattice dynamics data, Euphonic must be able to produce a function that takes a list of..."?

Would there actually be "function factory" code in the Euphonic package or would it sit in the interface on the Horace side? It's not something I'd expect to see in a generic library.

euphonic/design/03_design_document_v0.3.md
```diff
@@ -0,0 +1,531 @@
...
+be used to create most others. Force constants are calculated by an external
+code (e.g. CASTEP, Phonopy) and read to create this object. Force constants are
+useless on their own and require extra structure data (read in along with the
+force constants). The attributes required for this object are:
+
+|Attribute|Type|
+|---------|----|
+|Structure| Structure object|
+|Force constants|(n_cells, 3\*n_ions, 3\*n_ions) array of floats|
+|Cell origins|(n_cells, 3) array of ints|
```
ajjackson, line 172: I remember this briefly came up before, but I'm not sure it was seriously proposed/discussed.

Have we considered the option to:
- define and document a way of returning cell origins in a standard order from a given supercell matrix. (This could be as simple as "sort by value, in priority order R_a, R_b, R_c".)
- store the force constants matrix entries in that order
- remove this required attribute and replace it with a method that generates them as required from the supercell matrix?

Benefits of this approach:
- smaller, simpler object (and serial format)
- eliminate need to check that the list of cell origins is reasonable (i.e. all entries lie within supercell + no entries repeated + all origins in supercell accounted for).

Downsides:
- serial data is less self-explanatory
- data loaders may need to re-order the rows/columns of the force constants to match this scheme
- maybe costs a little performance (but it's not a big operation and results can be memoized if needed repeatedly)

euphonic/design/03_design_document_v0.3.md
```diff
@@ -0,0 +1,531 @@
...
+|Structure|Structure object|
+|Number of q-points|int|
+|Q-points|(n_qpts, 3) array of floats|
+|Q-point weights|(n_qpts,) array of floats|
+|Frequencies|(n_qpts, 3*n_ions) array of floats|
+|Eigenvectors|(n_qpts, 3*n_ions, n_ions, 3) array of complex floats|
+|Split frequencies| (n_gamma_points, 3*n_ions) array of floats|
+|Split eigenvectors|(n_gamma_points, 3*n_ions, n_ions, 3) array of complex floats|
+|Gamma point indices|(n_gamma_points,) array of ints| 
+|Energy unit|string|
```
ajjackson, line 206: This seems over-complicated. Why not just have a repeated value in the Q-points sequence, with corresponding entries in frequencies/eigenvectors?

euphonic/design/03_design_document_v0.3.md
```diff
@@ -0,0 +1,531 @@
...
+
+**Numpy Array** - Very simple, but wouldn't be able to bundle associated data
+together and would require the user to keep track of this. For serialisation,
+plotting, further calculations the user would have to pass all pieces of
+required data as arguments which is messy and likely to cause errors.
+
+## Units
+Units in Euphonic will be handled by the [`Pint`](https://pint.readthedocs.io)
+package. This provides a large list of unit conversions that will not have to be
+maintianed by Euphonic, and allows for addition of custom units. It handles
```
ajjackson: `maintianed`

euphonic/design/03_design_document_v0.3.md
```diff
@@ -0,0 +1,531 @@
...
+
+Plotting in Euphonic will be fairly simple, as it is not Euphonic's main goal,
+and will be mainly useful for debugging and quick sanity checking of answers. 
+
+## Serialisation
+For both easy testing and export to other programs, Euphonic will need to be
+able to serialise most of its data objects. They should each have their own
+methods, although it is not decided what formats should be supported and more
+may be added later. Initial support could be for one of: 
+
```
ajjackson, line 533: Are we still considering HDF5 for large objects?

ajjackson (commented):
Have taken a look through the design doc 03. Seems very agreeable!
- Abins overview is fine
- Slightly confused about Horace interface (commented in-place)
- Approach to CASTEP things seems healthy all-round
- Convincing case made for regular classes
- Useful overview of Pint mechanics
- Most objects seem good, have commented where I saw potential redundancy/complication

euphonic/design/03_design_document_v0.3.md
rebeccafair: I don't think we've decided on the Euphonic/Brille/Horace hierarchy yet, I think we shouldn't require Brille (as far as I know it has some relatively hefty dependencies), but we shouldn't preclude using it either. @g5t do you have any thoughts?

euphonic/design/03_design_document_v0.3.md
rebeccafair: I think we'd have a function on the Horace side that makes all the calls to Euphonic required to go from qpts -> frequencies + structure factors. I forgot to mention there's an example of that function [here](https://github.com/pace-neutrons/Meuphonic/blob/master/matlab/euphonic/euphonic_sf.m). The Python/Matlab interface there is terrible and I would definitely rewrite it, but you get the idea.

euphonic/design/03_design_document_v0.3.md
rebeccafair: Thanks, will add

euphonic/design/03_design_document_v0.3.md
rebeccafair: I feel like explicitly storing the cell origins is better, as it doesn't cost that much memory (compared to the force constants) and having the serial data be more self-explanatory is definitely a good thing.

Is there an easy way to generate cell origins for a skew cell? The supercell matrices I've been working with so far have all been diagonal.

ajjackson: In that case, as the data is semi-redundant, we should have sanity-checks on initialisation and on file-loading to make sure the set of supercell lattice points equals the stored set of origins. This way both of them can be safely relied on during use.

Generating origins consistently doesn't _look_ like a hard problem, but we will need a strategy to do it at some point so it's worth thinking about. (It doesn't need specifying in this document though!)

euphonic/design/03_design_document_v0.3.md
rebeccafair: This is a good point. I initially wrote it like this because I felt like if you pass in a list of `N` q-points, and the object that is returned has a list of `N + 1` q-points due to splitting that might be confusing. However, I've been having second thoughts about it and agree that it is overly complicated. If you think that having a repeated value is better and is the more 'usual' thing to do, I'm happy to do that instead, I know this is what CASTEP does at least.

euphonic/design/03_design_document_v0.3.md
rebeccafair: Yes, I think that's our most likely data format for large objects. I just listed some human-readable file types as that's probably what will get implemented first because I'll be using it for testing. I'll add a mention of HDF5.

euphonic/design/03_design_document_v0.3.md
g5t: Building the Python interface for `brille` requires:

1. a C++14 compliant compiler
2. CMake &ge; v3.13
3. pybind11 &ge; v2.2.0
4. [*optional but strongly recommended*] OpenMP &ge; v2.0

The first _shouldn't_ be difficult, but some linux systems have ancient system compilers (e.g., CentOS 7.7 uses gcc 4.8.5 which only has _experimental_ support for __*C++11*__).
To get around pre-C++-14 system compilers I was planning to build and distribute binaries on PyPI using [manylinux](github.com/pypa/manylinux). Thus far I haven't managed to build `brille` for `manylinux1` and the `manylinux2010` wheels I've made were ignored by `pip`.

At present CMake, pybind11 and OpenMP need to be installed system-wide, but it may be possible to reconfigure `brille`'s `setup.py` to use CMake and pybind11 from PyPI. If OpenMP isn't present `brille` should still compile and run.

If I sort out the manylinux issue and/or change the build to not rely on pre-installed system utilities/libraries then it would be simple to add `brille` as a dependency of `Euphonic`.

As to which hierarchy of

- `Horace` &rightarrow; `Euphonic` &rightarrow; `brille`
- `Horace` &rightarrow; `brille` &rightarrow; `Euphonic`

is desirable: ¯\\\_(ツ)\_/¯

rebeccafair: Ah, that's easier than I thought. I'll add some arrows to the context diagram and a mention this possibility in the Horace/Brille section. I can't think of anything in the design which would prevent us putting in an extra step to use Brille in the future.

euphonic/design/03_design_document_v0.3.md
ajjackson: I see, it is a little uncomfortable that the output data is not going to match the input sequence. Pushing some of the values into another attribute hides this problem rather than solves it. We should clearly establish that the usage pattern for this object is to work with the data it produces and not try to combine it with leftover input data.

Maybe drop a note in the document  here clarifying that n_qpts does not equal the number of _unique_ q-points. Although that conflates this issue with points that are revisited on a band-structure path; there must be a better way of wording it!

rebeccafair: I'll add a note on this

rebeccafair:
Given that this PR has been open for a while, I think I'm going to merge it before we start to overthink things. If anything comes up along the way we can still change things.

PR #5, opened 2020-03-04
Euphonic Validation Summary

rebeccafair:
I've added an initial Euphonic validation summary with placeholders for things we haven't decided yet (e.g. where to store large .sqw files).

I've written it on the assumption that we can only get the binned data from OClimax (as that's what we currently have access to), we still need to ask the OClimax developers if there's a way for us to get access to the mode resolved structure factors. In any case I think it might be a good idea to have some sort of validation on the binned data too, to show Euphonic is actually binning, applying the Bose population factor, and plotting correctly.

In the meeting we initially decided a combination of 1st BZ spaghetti plots & random points might be good for comparing with Ab2tds and OClimax, but we probably actually can't do this in Ab2tds/OClimax without significant effort, see [this section](https://github.com/pace-neutrons/pace-developers/blob/euphonic_validation_review/euphonic/validation/01_overview.md#comparison-with-software). So we need to come up with an alternative, I guess using a selection of the Horace q-cuts?

Also, the section on the actual metrics that will be used is also quite vague and should be expanded on.

Comments welcome.


krefson (approved):
This looks like a comprehensive description of the plan, and I could not see anything to change.

PR #7, opened 2020-03-06
Powder-averaging discussion

ajjackson:
Notes from the powder-averaging discussion of 5/3/20 with @mducle @g5t @krefson @rebeccafair (@tgperring excused). This progresses https://github.com/pace-neutrons/Pace-Project-Plan/issues/25

The intended outcome of the meeting was some design documentation and some new project issues. We should confirm the documentation output before writing the issues; at the bottom of these notes I have added a section outlining the suggested tasks.

I could re-write these notes into a more approachable/coherent explanation, but in the process they would lose some of their value as a capture of the discussion. I suggest therefore that with appropriate revisions these notes form a "discussion" document and stand-alone design documents are created as appropriate for the different code developments that come out of it. Open to alternative approaches. Otherwise, please suggest corrections/oversights and areas where more clarification should be inserted.

### DIFF
--- a/euphonic/design/05_powder_averaging_discussion.md
+++ b/euphonic/design/05_powder_averaging_discussion.md
@@ -4,9 +4,13 @@ These notes were initially drawn up by Adam Jackson. They were
 discussed and developed in a meeting on 5/3/20 with Greg Tucker, Duc
 Le, Keith Refson and Rebecca Fair.
 
+In this document **q** refers to a 3-dimensional momentum transfer
+vector, while *q* refers to "mod q", a scalar value of momentum
+transfer.
+
 # Table of Contents
 
-1.  [PACE project issues](#github-issues)
+1.  [Related Issue pages](#github-issues)
 2.  [Background](#org0814430)
 3.  [Analytic method](#org7f363ce)
 4.  [Numerical methods](#orgfd3e1f7)
@@ -25,17 +29,22 @@ Le, Keith Refson and Rebecca Fair.
 
 -   [#25 Euphonic Calculations: Powder averaging library](https://github.com/pace-neutrons/Pace-Project-Plan/issues/25)
     - The Issue directly leading to this discussion
--   Incoherent analytic overtones <https://github.com/pace-neutrons/Pace-Project-Plan/issues/24>
 
+- [#24 Euphonic Calculations: Overtones and Combinations (incoherent semi-analytic)](https://github.com/pace-neutrons/Pace-Project-Plan/issues/24)
+  - Higher-level issue for implementation of CLIMAX-like analytic powder averaging
+  - Leads to #67 (implementation in Euphonic) and #68 (refactoring Abins to use Euphonic)
 
 <a id="org0814430"></a>
 
 # Background
 
--   Single-crystal samples are typical on MAPS
+Powder averaging is appropriate to simulation of many neutron instrument measurements
+
+-   Single-crystal samples are typical on MAPS, but powder measurements are also performed
 -   Powdered crystalline samples are required on MARI
 -   Weakly-bound polycrystalline/liquid (cooled from liquid/gas?)
-    samples are typical on TOSCA (especially hydrogenous)
+    samples are typical on TOSCA. (Generally these are hydrogenous,
+    leading to spectra dominated by incoherent scattering.)
 -   There are many exceptions to these, so versatility is useful
 
 -   Currently Euphonic computes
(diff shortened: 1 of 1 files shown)


ajjackson:
The easiest way to view the document from this page is to go to the "Files Changed" tab
Screenshot from 2020-03-06 11-17-55

and then click the page icon to "display rich diff"
Screenshot from 2020-03-06 11-20-54

euphonic/design/05_powder_averaging_discussion.md
```diff
@@ -0,0 +1,318 @@
...
+-   There are two different possible objectives here: uniform sampling
+    **at each q** and uniform sampling **over the spherical
+    volume**. Which is more important?
+
+-   Why do we pre-determine the mod(q) bins?
+    -   easier to compare with instrument
+    -   able to check convergence *at each bin* independently
+    -   in Abins, the simulated *q* bins do not match the TOSCA
+        instrument measurements, in which the bin size increases with
+        *q*.
```
rebeccafair, line 251: If we were to randomly sample within each *q* bin, we could have arbitrary *q* bin widths which would enable us to better match the instrumental data. This would probably be much better at larger bin sizes, rather than assuming a single *q* (or a few *q*) accurately represent the instrumental measurements in that bin.

ajjackson: Right, this is a challenge with binning existing randomly-distributed data: some bins might be empty or have results heavily biased by an outlier. Using large bins reduces this problem but is not guaranteed to eliminate it.

euphonic/design/05_powder_averaging_discussion.md
```diff
@@ -0,0 +1,318 @@
...
+  while S reaches out further in reciprocal space. We'll probably stick with
+  Monkhorst-Pack grids initially.
+
+
+### Can we use statistical methods to quantify uncertainty in the convergence?
+
+- Statistical methods, propogation of error approaches could provide
+  some useful tools
+- Experimentally we can examine uncertainty by comparing different
+  detector contributions to the same bin
```
mducle: I guess you mean here that we would have multiple detectors at the same 2theta but different phi angles so they would all have the same abs(q) so contribute the same abs(q) bin?

If so, maybe you can change "same bin" to "same abs(q) bin" to make it clearer? (sorry it took me a few seconds to understand).

euphonic/design/05_powder_averaging_discussion.md
```diff
@@ -0,0 +1,318 @@
...
+
+
+<a id="orgce0cfe3"></a>
+
+### Random sampling
+
+-   This has been mentioned in PACE meetings. How much has it been
+    used?
+    - Random sampling on spherical shells seems more common that fully
+      random sampling in **q**-space.
```
mducle, line 189: An example is found in `spinw`: https://github.com/SpinW/spinw/blob/master/swfiles/%40spinw/powspec.m#L291 which does random sampling on specified abs(q) shells.

euphonic/design/05_powder_averaging_discussion.md
```diff
@@ -0,0 +1,318 @@
...
+
+
+<a id="orgd1b667c"></a>
+
+### Spherical sampling
+
+-   SpinW seems to use a Fibonacci-related method for uniform
+    sampling at spherical surfaces
+    -   points per shell has to be a Fibonacci number
+        - That's not a deal-breaker but could be annoying when exploring convergence
```
mducle, line 212: Code link: https://github.com/SpinW/spinw/blob/master/swfiles/%40spinw/powspec.m#L268

euphonic/design/05_powder_averaging_discussion.md
```diff
@@ -0,0 +1,318 @@
...
+    -   or can sample *randomly* at *each* spherical shell surface
+-   MDANSE requires the user to set up a q lattice for sampling,
+    including a "spherical grid" option with a set range of "shells"
+    and "vectors".
```
mducle, line 217: Not sure if this is similar to MDANSE, but `mcphase` does a very primitive *regular* spherical grid (specified by number of points along theta and phi directions and the by number of abs(q) bins): https://github.com/mducle/mcphase/blob/master/bin/powdermagnon.pl#L225 

This is *not* the best way!

euphonic/design/05_powder_averaging_discussion.md
```diff
@@ -0,0 +1,318 @@
...
+
+    -   Ideally one would build a combined output from every simulated pixel
+        of the instrument. This doesn't seem feasible for instruments
+        with thousands of detectors (e.g. MERLIN)
+        -   This approach seems to map to the McVine package's goals
+            better than Horace/Euphonic.
+        -   McVine accounts for instrument design, but takes S(q,ω) as
+            input; no support for 4-D data, which could be needed for such a model.
+        -   The McVine package is somewhat intimidating; such work is
+            not to be taken lightly.
```
mducle, line 273: [McStas](http://mcstas.org/) is also another package which can be used to do something similar: https://www.youtube.com/watch?v=ZcqermI64VA but if anything it can be even more complicated than McVine.

euphonic/design/05_powder_averaging_discussion.md
```diff
@@ -0,0 +1,318 @@
...
+High-priority independent work packages to be progressed are:
+
+- Analytic method for incoherent scattering systems
+- Spherical averaging on a requested sequence of *q* values, with options to
+  - Sample in a uniform way (e.g. Fibonacci scheme)
+  - Sample randomly
+
+Further scoping work should be performed for:
+
+- Sampling from an arbitrary collection points in **q**-space (not regular shells)
```
mducle:
```suggestion
- Sampling from an arbitrary collection of points in **q**-space (not regular shells)
```

mducle (approved):
Looks good to me.

euphonic/design/05_powder_averaging_discussion.md
ajjackson: I think that's probably what was meant.

But out of interest, would it also be possible to compare experimental detectors contributing to the same **q** on some instruments? Is there some overlap between detectors with a different 2(theta) at a different time-of-flight?

mducle: Yes (I think).

So, actually the ToF dependence of each detector represents a 1D path in (**q**, ω), so different detectors *do* contribute to the same **q**, but at different ω. 

In practice what we do with the data is to first convert from ToF to energy transfer ω, and then rebin it in a regular ω grid (the ToF data would probably have been measured in a regular ToF grid which would yield a ragged ω grid).

This then gives data where the **q** bins of each ω-binned strip would differ from ω-bin to ω-bin (because they intersect the 1D (**q**, ω) trajectory of the detectors at different points).

For single-crystal data, the data is stored on disk in this ragged-**q** bin format and rebinned into a regular **q** grid on the fly by the data analysis software (either MSlice or Horace). 

For polycrystalline data, the data is actually rebinned in regular *2-theta* bins (so ragged in abs(q)) and stored on disk in this grid. It is then rebinned into regular abs(q) bins by the data analysis software (MSlice).

To have different detectors with the same **q** *and* ω would be possible on instruments like MARI/MAPS/MERLIN but would need multiple measurements with different incident neutron energy/wavelength. It would not be possible on a TOSCA-like instrument because the scattered neutron energy is fixed by the analyser crystals and cannot be varied from measurement to measurement.

ajjackson:
Thank you Becky and Duc. I will merge this now but of course further revisions are possible if needed.

PR #8, opened 2020-03-23
Python interface design documents

mducle:
This is a description of prototyping work done to check how we can implement a Python interface to Matlab code (specifically Horace/Herbert), and more generally how Matlab and Python code can interact with each other.

There are two documents:

1. [A high-level description](https://github.com/pace-neutrons/pace-developers/blob/python_interface/python_interface/design/01_pace_python_high_level_discussion.md)
2. [A more detailed view of the prototype implementation](https://github.com/pace-neutrons/pace-developers/blob/python_interface/python_interface/design/02_pace_python_implementation_discussion.md)

The actual implementation is found [elsewhere](https://github.com/mducle/hugo), but is referred to extensively by these documents.

Associated issues:

https://github.com/pace-neutrons/Horace/issues/26
https://github.com/pace-neutrons/Pace-Project-Plan/issues/78


python_interface/design/01_pace_python_high_level_discussion.md
```diff
@@ -0,0 +1,182 @@
...
+
+# Scope
+
+This document summarises, at a high level, methods for running interdependent Matlab and Python code alongside each other, with and without a Matlab license.
+Implementation details are presented in a [separate document](02_pace_python_implementation_discussion.md).
+
+
+# Motivation
+
+PACE will be a suit of several interconnected programs for inelastic neutron data analysis, which are written both in Matlab and in Python.
```
g5t:
```suggestion
PACE will be a suite of several interconnected programs for inelastic neutron scattering data analysis, which are written both in Matlab and in Python.
```

nickbattam-tessella:
Are the limitations on which releases of MATLAB and the MCR this model will work against? `Hugo` is currently configured to run with R2019a / MCRv96.

Horace is being targeted/built/tested against R2018a and later. For builds targeted to integration with Python there is no requirement for us to support multiple MCR versions. We build create a single compiled release against any specified R20xx independently of the multi-version "MATLAB releases (with mex files)" but will need to consider whether we support one or many.

mducle:
@nickbattam-tessella There are no restrictions - R2019a is just what I have installed at the moment. It should work for any Matlab newer than 2018a. The catch is that users have to download the MCR version against which we compile (or we bundle the MCR with the distribution).

nickbattam-tessella:
> The catch is that users have to download the MCR version against which we compile (or we bundle the MCR with the distribution).

Unfortunately requiring the user to download the MCR is necessary as I understand the documentation (https://uk.mathworks.com/help/compiler/deployment-process.html): 
```
The installers generated by the compiler apps may include the MATLAB Runtime installer. If you
compiled your artifact using mcc, you should direct your end-users to download the MATLAB
Runtime installer from the website https://www.mathworks.com/products/compiler/mcr.
```
We will need to clearly document the version they *must* use in the release notes, a read.me, ....

mducle:
@nickbattam-tessella - so I didn't emphasise in the doc, but we *can* use `deploytool` as a commandline to generate an installer, using the [`-package`](https://uk.mathworks.com/help/compiler_sdk/ml_code/deploytool.html) option. I'll change the relevant parts in the doc.

The commandline functionality of `deploytool` requires an XML input file (with a `.prj` extension) which can be generated by the GUI version, but we can also generate it programmatically. In this `.prj` file is a boolean option `param.package.mcr` which specifies whether you want to include the MCR inside the installer (so users don't have to download it). In the GUI, this option is in the top strip panel and labelled "Runtime included in package".

Apparently, from R2020a, we can also use a commandline [`compiler.package.installer`](https://uk.mathworks.com/help/compiler/compiler.package.installer.html) command with an option `'RuntimeDelivery'` to do this instead of using the `.prj` file.

mducle:
I don't really want to package the MCR with an installer though - for one thing, we would have to host >1GB packages - but just wanted to note that it is possible. Ideally I want to be able to package it for installation via `pip` or `conda` which might not allow such large packages.

python_interface/design/01_pace_python_high_level_discussion.md
```diff
@@ -0,0 +1,182 @@
...
+An alternative which would use only pure Matlab or Python code (without needing any compilation) is to use a [reverse-communication interface](#reverse-communication-interface).
+The implications of these proposed implementations on parallelisation is discussed in the [final section](#parallelisation).
+We end with a [summary and discussion points](#summary).
+
+
+# <a name="python-interface"></a> Python User Interface
+
+As detailed in [this section](02_pace_python_implementation_discussion.md#compiled_matlab), Matlab code may be "compiled" using the **Matlab Compiler** and freely distributed to users without a Matlab license.
+In order to run this "compiled" code, the user is required to install the free/gratis **Matlab Compiler Runtime** package, and this **MCR** package must match the version of Matlab used to compile the Horace library.
+Thus, if we always target the latest version of Matlab, we should require users to re-install (upgrade) the **MCR** twice a year, which may not be desirable.
```
nickbattam-tessella: I'm not sure there is any benefit in following the latest MATLAB releases. The compiled route is a self-contained system so we can upgrade when it is beneficial to do so (new features, bug fixes in the MCR, ...).

python_interface/design/01_pace_python_high_level_discussion.md
```diff
@@ -0,0 +1,182 @@
...
+
+The compiled Matlab library is a platform-independent Component Technology File (`ctf`) which contains compressed and encrypted versions of the `m` and `mex`-files used in the project.
+The **MCR** is a platform-dependent installation (approximately 1GB download) which contains the base Matlab distribution and libraries but is set up to be able to run only code from `ctf`s.
```
nickbattam-tessella: Would it be helpful to add a link to the MCR download site: https://uk.mathworks.com/products/compiler/matlab-runtime.html

python_interface/design/01_pace_python_high_level_discussion.md
```diff
@@ -0,0 +1,182 @@
...
+This has implications for calling Python functions from Matlab using the Python interface, as discussed in the next section.
+
+Since array data within a `matlab.object` is *not* copied when exported from Matlab to Python, we may use it to wrap large data arrays, to avoid the data copying discussed in the previous section.
+However, the reverse direction is more problematic since the Matlab-Python library does not allow arbitrary Python objects to be exported to Matlab.
+Thus large array exports from Python to Matlab will need either to be copied using the default export mechanism, or we need to implement the shared-data [export mechanism described in the details](02_pace_python_implementation_discussion.md#mex_wrapping).
+This will definitely be needed if large parts of PACE come to be written in Python, and only small parts remain in Matlab and will thus need to communicate much more data with Python.
+As a bonus feature, the share-copies mechanism will allow users fast access to the `pix` array of an `sqw` object for example.
```
nickbattam-tessella, line 94: The plan for Horace was to change the primary interaction with the pixel array from direct array access to column-based getters and setter, i.e. rather than accessing `sqw.data.pix[6]` the user should call `sqw.data.getSignal()` or `sqw.data.getData('signal')`.
Does that design decision have implications on here?

python_interface/design/01_pace_python_high_level_discussion.md
```diff
@@ -0,0 +1,182 @@
...
+w2 = w1.cut([0.1, 0.5], [3, 0.5, 20])   # or m.cut(w1, [0.1, 0.5], [3, 0.5, 20])
+w2.plot()                               # or m.plot(w2)
+```
+
+Note that it is not possible to import all Matlab functions or classes into the Python namespace directly so syntax which omits the `m.` will only be possible by explicitly specifying individual Python wrappers to equivalent Matlab functions.
+
+
+# <a name="interpreter-calls"></a> Calling user defined functions
+
+Depending on the outcome of the design review for the optimiser, described [here](/optimisation/design/Model_Optimisation_Design.md), the optimiser may be written in Matlab or in Python (or perhaps C++).
```
nickbattam-tessella, line 114: You look at the MATLAB and Python optimiser use-cases. How would a pure C++ optimiser be used?

python_interface/design/02_pace_python_implementation_discussion.md
```diff
@@ -0,0 +1,349 @@
...
+`mcc` only parses `m` and `mex` files, so, because the `Horace` and `Herbert` distributions contains many files which are not `m` nor `mex`, the build process copies only the `m`/`mex` files to a new folder structure.
+`mcc` gives an error on zero-length files and on the markup tags used by `docify` so these are removed from the temporary folder using a shell script.
+A single `m`- or `mex`-file has to be specified as the target of the compilation, which is `pyhorace_init.m` in our case.
+The `-a` switch then adds an `m`- or `mex`-file or a folder to the `ctf`, and may be specified repeatedly. 
+Each `m`- or `mex`-file specified by an `-a` switch is available to be called directly by Python.
+`m`- or `mex`-files contained within a folder specified by `-a` are included in the `ctf` archive but cannot be called directly by Python.
+For `pyHorace`, two `m`-files which internally uses `feval` are specified by the `-a` switch (`call.m` and `call2.m`) to call the actual Horace/Herbert routines which are included only as folders.
+`call.m` is to call functions directly, whilst `call2.m` calls methods of objects.
```
nickbattam-tessella: `call2` is not a clear/descriptive method name. I don't have any clues from the how this differs from `call`.

python_interface/design/02_pace_python_implementation_discussion.md
```diff
@@ -0,0 +1,349 @@
...
+Thus, if we were to pass a `numpy` array (which is already a type which follows the `buffer` protocol and so is convertible to `array.array` without copying) we need to convert it to a `list` which involves an *additional* data copy.
+Thus the default method to pass data from a `numpy` array to Matlab requires *three* data copies (`numpy` to `list`, `list` to `buffer`, and internally within Matlab).
+The [prototype implementation](https://github.com/mducle/hugo/blob/master/pyHorace/TypeWrappers.py#L25) avoids the two unecessary Python-side copies by constructing a `matlab.<type>` object which directly wraps the `numpy` array.
+It can only do this for `numpy` arrays which are stored in memory in a column-major (Fortran-style) layout, however, because the Matlab-Python library only supports such a layout.
+(There are indications that internally the `mxArray` *can* support a row-major layout, but this has not been explored in detail).
+In any case, for row-major (C-style) `numpy` arrays, the prototype implementation is forced to perform a data copy.
+
+## Complex-valued arrays
```
nickbattam-tessella, line 96: Out of curiosity are there many complex-values arrays in the fix/optimisation algorithms that will need need moving between Python and MATLAB?

python_interface/design/02_pace_python_implementation_discussion.md
```diff
@@ -0,0 +1,349 @@
...
+6. When it is called, `call_python.mex` then checks the `dict` in `pyHorace.FunctionWrapper` to see if the key is valid and if it is, calls the function referenced by it.
+7. `call_python.mex` wraps `float`/`double` and `complex<float>`/`complex<double>` Matlab arrays in a `numpy` array wrapper before passing these to the Python function (other wrappers are possible but have not been implemented).
+8. On return, `call_python.mex` copies the data in the `numpy` array (the prototype implementation assumes only one array is returned) into a Matlab array and returns this to Matlab.
+
+The reason to wrap the function key in a `pythonFunctionWrapper` object is to ensure that the function reference does not persist eternally in the `FunctionWrapper` `dict`, causing a memory leak.
+When the `pythonFunctionWrapper` object goes out of scope, it should be deleted by Matlab, and this calls its destructor `mex` function to remove the reference from the global `dict`.
+
+As noted previously, the `thinwrapper` used by the prototype implementation does not handle nested old-style classes, such as the `multifit_sqw` class which has a nested `sqw` object within it.
```
nickbattam-tessella, line 185: the old-style classes are being actively removed from the Horace/Herbert code so this shouldn't be a problem.

python_interface/design/02_pace_python_implementation_discussion.md
```diff
@@ -0,0 +1,349 @@
...
+
+```python
+import struct
+import ctypes
+
+from pyHorace import Matlab
+m = Matlab()
+
+# We cannot directly create a variable in the Matlab base namespace
+# So we first create it in Python using the default matlab.double ctor and get its memory address
```
nickbattam-tessella: `ctor`? Unnecessary abbreviation

python_interface/design/02_pace_python_implementation_discussion.md
```diff
@@ -0,0 +1,349 @@
...
+print(type(mm_py._data))
+print(mm_py._data)
+# The address of the buffer can be obtained from the buffer_info property
+# The first element is the address, the second is the length
+buf_info = mm_py._data.buffer_info()
+print(buf_info)
+print('Exported Python array memory address is {}'.format(buf_info[0]))
+
+# Check that the memory pointed to by the buffer info stores the right data
+# Assumming the data is stored as IEEE double, 64-bit (8 byte) floating point number
```
nickbattam-tessella, line 252: Herbert/Horace is using a mix of IEEE float (32-bit) and doubles (64-bit) for data. Specifically the pixel data signal/error data is stored as 32-bit floats (@abuts may be able to add more specific details).
This interface between Python and MATLAB will need to be able to handle both float and double data.

python_interface/design/01_pace_python_high_level_discussion.md
mducle:
```suggestion
Thus, it would make sense to target a specific Matlab release for stability until new features make upgrading desirable.
```

python_interface/design/01_pace_python_high_level_discussion.md
mducle:
```suggestion
The **MCR** is a platform-dependent installation ([approximately 1GB download](https://uk.mathworks.com/products/compiler/matlab-runtime.html)) which contains the base Matlab distribution and libraries but is set up to be able to run only code from `ctf`s.
```

python_interface/design/01_pace_python_high_level_discussion.md
mducle: I'm not sure - in principle no. I think the Python interface will respect `Access = private` attributes on Matlab objects so this will not be visible to Python users either, just the setter/getter; but I need to check. The getter/setter will be accessed in Python just the same as in Matlab.

python_interface/design/01_pace_python_high_level_discussion.md
mducle: It would need a wrapper each for the Python (probably using PyBind11) and Matlab interfaces (using either mex or if we decide to use Matlab R2019a or newer, the `clibgen` functions to export a set of headers so Matlab can call the C++ library directly - see https://blogs.mathworks.com/developer/2019/07/11/cpp-interface/ ).

For data exchange, on the Matlab <-> C++ front, we would probably have to write mex functions similar to what would be needed if the optimiser was written in Python (albeit probably simpler than if it was in Python) to ensure that the data was not needlessly copied. I don't know how `clibgen` works in this respects though. On the Python <-> C++ front, PyBind11 does all the heavy lifting so it would be even simpler.

python_interface/design/02_pace_python_implementation_discussion.md
```diff
@@ -0,0 +1,349 @@
...
+[An example CMake](https://github.com/mducle/hugo/blob/master/src/CMakeLists.txt#L49) file shows how this could be implemented.
+`mcc` only parses `m` and `mex` files, so, because the `Horace` and `Herbert` distributions contains many files which are not `m` nor `mex`, the build process copies only the `m`/`mex` files to a new folder structure.
+`mcc` gives an error on zero-length files and on the markup tags used by `docify` so these are removed from the temporary folder using a shell script.
+A single `m`- or `mex`-file has to be specified as the target of the compilation, which is `pyhorace_init.m` in our case.
+The `-a` switch then adds an `m`- or `mex`-file or a folder to the `ctf`, and may be specified repeatedly. 
+Each `m`- or `mex`-file specified by an `-a` switch is available to be called directly by Python.
+`m`- or `mex`-files contained within a folder specified by `-a` are included in the `ctf` archive but cannot be called directly by Python.
+For `pyHorace`, two `m`-files which internally uses `feval` are specified by the `-a` switch (`call.m` and `call2.m`) to call the actual Horace/Herbert routines which are included only as folders.
```
mducle:
```suggestion
For `pyHorace`, two `m`-files which internally uses `feval` are specified by the `-a` switch (`call.m` and `call_method.m`) to call the actual Horace/Herbert routines which are included only as folders.
```

python_interface/design/02_pace_python_implementation_discussion.md
mducle:
```suggestion
`call.m` is to call functions directly, whilst `call_method.m` calls methods of objects.
```

I'll also change the code in `hugo`.

python_interface/design/02_pace_python_implementation_discussion.md
mducle: I don't think there are any... for the optimiser anyway (it should only be passing coordinates, parameters and intensities/energies around, and these should all be real quantities). 

I guess I was trying to be general, and in other contexts we might need to pass complex arrays around (in a non-Horace context, for pySpinW it would need to pass complex arrays back and forth).

python_interface/design/02_pace_python_implementation_discussion.md
mducle:
```suggestion
# So we first create it in Python using the default matlab.double constructor and get its memory address
```

python_interface/design/02_pace_python_implementation_discussion.md
mducle: It should be able to do this - the interface `mex` functions will use templates to satisfy the full list of types. I just needed to assume one type for the `struct.unpack` call in the next line (since this is Python looking at raw memory)

mducle (commented):
Suggestions to autopatch.

abuts (commented):
Can not say I fully understand this but can I summarize what I understand as follows:
1) We can run compiled Matlab objects through Pyhon and apply various methods of Matlab objects. 
2) Exchange of all small data between Python and Matlab is not a problem using copying and there is extensive API for doing that. 
3) Python allows to initiate an numpy array accepting a pointer to correct memory location (technical constrains on valid memory layout applied)
4) One can receive `const * const` pointer to large data from Matlab and set it up as read-only input to Python (casting would not work as recent Matlabs checks the integrity of this array)
5) One can allocate memory, Matlab is responsible for and return it to Matlab using mex file.
6) 5-means that if python needs to transfer data to Matlab, it needs to ask special allocation interface from Matlab (written as a mex wrapper) which would return pointer to a Matlab's allocated and writeable memory, set up numpy array from this pointer, fill this memory up by python function results and then call Matlab's object method, which would utilize this memory.

Looks like a working strategy if all this is right. But user would need to call special command whithin a user functuion. 

7) One can write speciall wrapper which would allow Matlab calling user function in Python, so wrapper would register and Matlab objects would call the function, but it still should be function written according to the specific rules (pretty simple rules, bit instead of, say, `to_fill_np= numpy.array()` one need to call e.g. `to_fill_np, return_to_matlab=matlab_horace.array()` within the user function to return results of user function to Matlab where `to_fill_np` and `return_to_matlab` would be different types of wrappers around the same array (and `return_to_matlab` may be hidden from users, while `matlab_horace.array()` probably not)

mducle:
@abuts on points 5-7, I was hoping to be able to do something more convenient for the user, where the idea would be that going from Python to Matlab,

1. I would make a C++ `std::unique_ptr` to the data in a numpy array (in doing this I would deallocate the numpy _object_ so that it does not try to free the memory associated with the unique pointer).
2. Pass this unique pointer to the Matlab `createArrayFromBuffer` function using `std::move` which it accepts. This transfers ownership of the memory to Matlab.

On the first point, I can't find a way for Python to release the memory. The work-around I've come to is to manually change the `OWNDATA` flag in the numpy object from `true` to `false` so it doesn't try to delete the memory when the numpy object is deleted.

On the second point, the construction of the array works ok, but deleting the object causes a memory error. This is because the Matlab module and the Python interpreter were compiled using different compilers, they have different heaps. The data in memory was originally allocated by Python on one heap, but when Matlab tries to deallocate (using its heap, which is different), we get a heap memory error.eap).

In principle the `std::unique_ptr` has an associated `deleter` function pointer - I tried to set this to a function which does nothing (e.g. does not try to free the data memory block when the object is deleted; and then to manage all memory on the Python side), but the `createArrayFromBuffer` function overwrites this with its own deleter function.

The only way around this that I can think of is to create shared-data (shallow) copies in Matlab of all exported arrays and to put these copies into a cell array in a singleton global class in the base namespace, which are never freed. This ensures that Matlab would never try to free memory created by Python, but will cause a small memory leak due to all the wrapper objects. And it's an ugly solution...

PR #9, opened 2020-04-08
Resolution convolution design for PACE

g5t:
Addressing https://github.com/pace-neutrons/Pace-Project-Plan/issues/5
As described in detail in the design document, the plan would be to add an empty `initialise()` method to all `OptFunction` objects and then introduce new subclasses of `OptFunction` to implement resolution convolution.


resolution/design/00_resolution.md
```diff
@@ -0,0 +1,331 @@
...
+and the energy transferred is the difference of the initial and final kinetic energy of the neutron
+
+![E = \frac{\hbar^2}{2m_\text{n}}(k_\text{i}^2 - k_\text{f}^2)]
+
+Neutron scattering instruments are typically constructed to allow an appreciable uncertainty in
+the initial and final momenta of detected neutrons, ![\delta\mathbf{k}_\text{i}] and ![\delta\mathbf{k}_\text{f}], respectively.
+These uncertainties naturally give rise to uncertainties in the momentum, ![\delta\mathbf{Q}], and energy, ![\delta E], transferred to the sample;
+and therefore uncertainties in the doubly-differential cross section.
+In the interest of simplified notation we can define a momentum-energy four vector ![\mathbb{Q} \equiv (\mathb{Q},E)], with uncertainty ![\delta\mathbb{Q}].
+The uncertainty in momentum-energy at a given ideal momentum-energy point gives rise to the instrumental resolution function ![R(\mathbb{X};\mathbb{Q})].
```
mducle: I think you need to state here that `\mathbb{X}` is another (set of) momentum-energy 4-coordinate(s) around the ideal coordinate `\mathbb{Q}` (which is the coordinate of the detector assuming a delta-function resolution [perfectly collimated beam, point sample size, zero-width moderator time]).

Alternatively, I think the notation `R(\mathbb{Q} - \mathbb{Q}_0)` where `\mathbb{Q}_0` is the ideal momentum-energy point is maybe a bit clearer? So then the variances are:

```
V_i(\mathbb{Q}_0) \equiv 
   \frac{\int d\mathbb{Q} Q_i^2 R(\mathbb{Q}-\mathbb{Q}_0)}{\int d\mathbb{Q} R(\mathbb{Q}-\mathbb{Q}_0)} - 
   \left(\frac{\int d\mathbb{Q} Q_i R(\mathbb{Q}-\mathbb{Q}_0)}{\int d\mathbb{Q} R(\mathbb{Q}-\mathbb{Q}_0)}\right)^2
```

with `\sqrt{\mathbb{V}(\mathbb{Q}_0)} \propto \delta\mathbb{Q}(\mathbb{Q}_0)` and the final integral below becomes 

```
R\left\{S(\mathbb{Q}_0;\mathbf{p})\right\} \equiv \int R(\mathbb{Q}-\mathbb{Q}_0) S(\mathbb{Q}) d\mathbb{Q}
```

Or is this wrong?

resolution/design/00_resolution.md
```diff
@@ -0,0 +1,331 @@
...
+
+# Context
+In line with the model optimisation provided by PACE,
+users will provide a function or class method which calculates and returns ![S(\mathbb{Q};\mathbf{p})]
+at an array of ![\mathbb{Q}] values and constant model parameter values ![\mathbf{p}].
+These user-provided scattering functions may account for some instrumental effects, e.g., slowly-varying backgrounds,
+and are unlikely to account for resolution.
+In cases where features of the model are of comparable size to the instrumental resolution an accurate estimate to model parameters will only be possible if the model is convoluted with the resolution _before_ comparison with measured intensity.
+
+By providing the necessary resolution wrapper to user-provided ![S(\mathbb{Q};\mathbf{p})] models
```
mducle:
```suggestion
By providing the necessary resolution wrapper to user-provided ![S(\mathbb{Q};\mathbf{p})] models,
```

resolution/design/00_resolution.md
```diff
@@ -0,0 +1,331 @@
...
+
+## Deliverables
+- modifications to the `OptModel`, `OptFunctions`, and `OptFunction` classes
+- a new `OptFunction` subclass, `TobyfitOptFunction` to perform resolution convolution following the `tobyfit` scheme
+- possibly more `OptFunction` subclasses to perform resolution convolution following other schemes
+
+
+# Existing Solution
+Resolution convolution is accomplished within the existing `multifit` framework through specialisation of the MATLAB class `mfclass`.
+Like [`OptModel`] `mfclass` contains independent-variable array(s), foreground and background function(s) plus their parameters and meta information for fixing and binding-together parameters.
```
mducle:
```suggestion
Like [`OptModel`], `mfclass` contains independent-variable array(s), foreground and background function(s) plus their parameters and meta information for fixing and binding-together parameters.
```

resolution/design/00_resolution.md
```diff
@@ -0,0 +1,331 @@
...
+obj = obj.set_fun(@user_sqw_function, p_i); % define a foreground function
+obj = obj.set_bfun(@user_bkg_function, p_j); % define a background function
+obj = obj.set_free(...); % optionally fix foreground parameters
+obj = obj.set_bfree(...); % optionally fix background parameters
+obj = obj.set_bind(...); % optionally bind foreground parameters
+obj = obj.set_bbind(...); % optionally bind background parameters
+[sim_obj(s), p_fit] = obj.fit();  
+```
+
+The flexibility provided by the interaction of `mfclass`, `mfclass_tobyfit`, `mfclass_wrapfun`, and `mfclass_plist` comes with the significant drawback that the code is difficult to understand and therefore difficult to maintain.
```
mducle: I still don't understand fully the hierarchy here (sorry I haven't looked at the code much). So, I understand that:

1. `mfclass_tobyfit` is a subclass of `mfclass` (not `mfclass_wrapfun`).
2. An `mfclass_wrapfun` instance is a (optional) property of `mfclass`.
3. Instances of `mfclass_plist` are properties of `mfclass_wrapfun`.
4. `mfclass_plist` can itself have instances of `mfclass_plist` as properties.

So, my questions are:

1. Does `mfclass_tobyfit` just overload the default `mfclass_wrapfun` instance in `mfclass`? (If so, why can't we just change the `wrapfun_` instance from the default to `tobyfit_wrapfun` (or whatever)? I guess it's the same reason you discarded the `resolution_scheme` property idea from the first commit?). Or does `mfclass_tobyfit` overload other parts of `mfclass` too? (If so why have `mfclass_wrapfun`?)
2. What are the `mfclass_plist` used for? (From the name, I would guess that they are parameter lists which have the extra parameters needed for the resolution convolution in addition to the user-model parameters). And why does the `mfclass_plist` class have an instance of its class as a property? (is it to support recursively wrapping functions - e.g. wrap the user function in a resolution function, and then wrap this resolution function in another function to calculate e.g. multiple scattering [I'm not sure how this would works...]?)

resolution/design/00_resolution.md
```diff
@@ -0,0 +1,331 @@
...
+
+Additionally, we have investigated new methods for performing the resolution convolution -- notably, reuse of in-resolution function evaluations for nearby detector pixels.
+
+In order to support multiple ways of estimating the instrument resolution and performing the convolution we require a flexible resolution convolution scheme to be implemented in conjunction with `OptFunction`.
+
+## `OptFunction` subclasses
+`OptFunction` will be extended to include an `initialise()` method which performs no action by default, and a `simulate()` method which calls the `initialise()` method before the `evaluate()` method. Same-named similar extensions will be added to `OptFunctions` as well.
+The two 'gateway' methods, `simulate()` and `optimise()`, in `OptModel` will be modified to first call its contained `OptFunctions` objects' `initialise()` and then to continue utilising a private `evaluate()` method to finish simulating or optimising the model.
+The `evaluate()` method of `OptFunction` and `OptFunctions` must remain public.
```
mducle: Do we need a `simulate()` method for `OptFunction`/`OptFunctions`? 

It seems to be just a convenience for users since `initialise()` and `evaluate()` are both public...

Alternatively can we use Matlab's [class access list](https://uk.mathworks.com/help/matlab/matlab_oop/selective-access-to-class-methods.html#btb03uh) to have something like C++'s `friend` classes? This would allow `initialise()` and `evaluate()` to be private in `OptFunction`/`OptFunctions` too (so the user only ever sees `simulate()` [and `optimise()` for `OptModel`]).

resolution/design/00_resolution.md
```diff
@@ -0,0 +1,331 @@
...
+
+With these changes we can then impelemnt a new subclass of `OptFunction` for any resolution convolution scheme.
```
mducle:
```suggestion
With these changes we can then implement a new subclass of `OptFunction` for any resolution convolution scheme.
```

resolution/design/00_resolution.md
```diff
@@ -0,0 +1,331 @@
...
+classdef OptModel
+    properties
+        multipliers {mustBeOptFunctions} = [];
+        foregrounds {mustBeOptFunctions} = [];
+        backgrounds {mustBeOptFunctions} = [];
+    end
+    methods
+        function varargout = simulate(obj, varargin)
+            obj.initialise();
+            varargout = obj.evaluate(varargin);
```
mducle: Should it be `varargin{:}` in the `evaluate` calls?

resolution/design/00_resolution.md
```diff
@@ -0,0 +1,331 @@
...
+
+% Create the OptModel object
+%   Note: without a specified 'multiplier' or 'background' sensible defaults are
+%         provided, namely @(x)1 and @(x)0, respectively.
+opt = OptModel([sqw_object_1, sqw_object_2], 'foreground', fg_funcs);
+```
+
+# Converting `mfclass_tobyfit` to `TobyfitOptFunction`
+Some of the functionality of `tobyfit`, e.g., fitting crystal orientation or moderator lineshape parameters, is implemented in `mfclass_tobyfit` by setting a property of the class to be a `struct` with a documented set of fields.
+If the `struct` is set then a number of parameters are appended to the standard user-function parameters which are stripped-off again before the user-function is evaluated.
```
mducle, line 280: Is this where the `mfclass_plist` come in?

mducle (changes requested):
Looks good, but I've got some questions (in the individual comments).

resolution/design/00_resolution.md
g5t: I'll extend this part of the document and introduce the covariance matrix, V<sub>ij</sub>(**Q**,E), since this is what is used extensively to approximate the resolution of, e.g., triple-axis neutron spectrometers (c.f, [ResLib](https://neutron.ethz.ch/research/resources/reslib.html)).

The notation I'm using here is probably not ideal, but the idea that I'm attempting to convey is that the resolution *function* is defined for a specific instrument configuration [which is characterised by a single (**Q**,E) point] and is valid over all momentum-energy space.
The idea in the notation, as with model functions S(**Q**,E; **p**), that that properties which *define* the continuous function are listed following a semicolon.
The resolution function at a specific point, say (**Q**<sub>0</sub>,E<sub>0</sub>), is defined by the instrument configuration and is continuous over all (**Q**,E) &mdash; so, R(**Q**,E; **Q**<sub>0</sub>,E<sub>0</sub>).
We could stick with exactly the same symbols for defining the covariance, but then would specify V<sub>ij</sub>(**Q**<sub>0</sub>,E<sub>0</sub>), when my goal was to emphasize that it too is continuous in the experimentally accessible part of (**Q**,E) space.

As for why this document does not use R(**Q** - **Q**<sub>0</sub>, E - E<sub>0</sub>;  **Q**<sub>0</sub>,E<sub>0</sub>): Such a construction is useful when you define the instrumental resolution function as a hyperellipsoid as it simplifies the 4-D Gaussian expression, but as such a definition is beyond the scope of this document (and isn't necessary for all resolution models) I've decided against introducing an origin shift in the integrals.

mducle: Thanks for the clarification.

resolution/design/00_resolution.md
g5t: There is a mistake in the class diagram I produced which has, unfortunately, caused some of your confusion &mdash; `mfclass_plist` does not contain another `mfclass_plist` as its property `plist_`.
Instead `plist_` is the input provided to the class constructor (initially) and can be 'wrapped' to include additional functions and parameters (?). I'm not entirely sure what it is intended for though, perhaps @tgperring can help us both out.

As for your first question:
`mfclass_tobyfit` extends the properties of `mfclass` to include *some* of the information needed to perform the resolution estimate and convolution. I do not believe that any methods of `mfclass` are modified by `mfclass_tobyfit`. Using the `mfclass_tobyfit` constructor directly would not yield a useful convolution object, however, since one must *also* define a special `mfclass_wrapfun` to actually do everything.
The `tobyfit()` gateway function constructs this special `mfclass_wrapfun` object by providing it with a resolution-estimate function and convolution function, both of which are implemented as extension methods for `sqw`. The exact functions provided to the `mfclass_wrapfun` constructor are dependent on the instrument type encoded in the `sqw` objects input to `tobyfit()`.

I can try and expand on this in the design document, but I'm not sure that I can do so in a concise manner.

mducle: Maybe a short sentence saying that the actual implementation is more complicated and needs overloading both `mfclass` and `mfclass_wrapfun`? (and emphasise that the new proposed design would only involve subclassing a single class?)

resolution/design/00_resolution.md
g5t: I was not aware of class access lists in MATLAB, but I am very happy to learn of the feature.

`OptFunction` and `OptFunctions` need to have a `simulate()` method in case `initialise()` is not an empty function *and* a user wants to evaluate their (possibly resolution convoluted) function or functions-with-a-reduction-operator *without* worrying about setting up everything for optimising the function/functions to measured data.

Removing `initialise()` and `evaluate()` from their public methods is certainly a good idea.
I'll modify the document to reflect this.

resolution/design/00_resolution.md
g5t: `mfclass_wrapfun` isn't overloaded, which I think contributes to the difficulty in understanding how `tobyfit()` works.

I hope that by removing `mfclass_wrapfun` and sticking everything in the subclasses of `OptFunction`, e.g., `TobyfitOptFunction`, that following the execution path of the code will be less confusing to new users/developers.
If the user-provided function needs to be 'wrapped', as in `tobyfit`, that should be explicit in the class which implements the `tobyfit` method of resolution convolution.

resolution/design/00_resolution.md
g5t: It's untested pseudo-code. But yes, I think you're right.

mducle: "brain-compiled" is a phrase I read somewhere on Stack-overflow.

rebeccafair (approved):
This all looks sensible to me.

mducle (approved):
Looks good to me now.

PR #10, opened 2020-05-18
Third party API design document

mducle:
This adds a design document for the API for user-defined model functions in PACE/Horace in multiple languages.

This design document constitutes one of the milestone of the PACE project, with a deadline of May 2020.

Addresses https://github.com/pace-neutrons/Pace-Project-Plan/issues/35

### DIFF
--- a/optimisation/design/Third_Party_API_Design.md
+++ b/optimisation/design/Third_Party_API_Design.md
@@ -11,8 +11,7 @@ and how they interface with Horace, using either Matlab, Python, C, C++ or Fortr
 # Motivation
 
 Analysis of inelastic neutron scattering data almost always requires a parameterised model of the measured scattering intensities.
-Horace, as currently implemented, requires this model to be in the form of a Matlab function.
-This function is required to either 
+Horace, as *currently implemented*, requires this model to be in the form of a Matlab function, which is required to either 
 
 1. Take a 4D vector of momentum-energy ![\mathbb{Q} \equiv (\mathb{Q},E)] and yield the scattering function ![S(\mathbb{Q})].
 2. Take the 3D vector of momentum transfer ![\mathbf{Q}] only and yield ![N] sets of vectors ![(E_n, S(\mathbf{Q}, E_n))] where ![N] is the number of dispersion branches.
@@ -34,8 +33,11 @@ In the current Horace implementation, these model functions are used by the `mul
 The [new design for PACE](Model_Optimisation_Design.md) proposes instead to use a host of classes including `OptFunction` and `OptModel`.
 `OptModel` will be more like the current `multifit` (or `tobyfit`) class,
 whilst `OptFunction` is a wrapper around the user supplied model function to handle parameter bindings and bounds currently handled by `multifit`.
-`OptFunction` should therefore have functionality to transform type 2 input to type 1 input
-whilst `OptModel` would expect only type 1 input.
+In this design, `OptModel` will expect its input (of type `OptFunction`) to behave like the type 1 functions listed above,
+whilst for other types of user functions, `OptFunction` will transform the input and outputs to the form that `OptModel` expects.
+In addition to the two types noted above which Horace currently handles, the [new API](Model_Optimisation_Design.md#calling-the-function-inside-optfunction)
+defines several additional standard function types (including an alternative to type 2 which also yields an intrinsic energy width for each mode), which it will automatically transform.
+For other functions, users would need to define the transformation themselves. 
 
 The current implementation of Horace is Matlab-focussed and user defined model functions must be Matlab functions.
 PACE, however, will have a Python version or interface.
@@ -55,8 +57,10 @@ This is because Matlab does not allow its own objects (instances of Matlab-defin
 (A workaround for this for the Python user interface using a compiled Horace library exists,
 and is described [here](../../python_interface/design/01_pace_python_high_level_discussion.md#interpreter-calls)).
 This means that the second type of function signatures described in the [optimisation design document](Model_Optimisation_Design.md#optimisation-function-object), 
+`array_objects(a, ..., p)`, `S_of_Q_w(sqw_obj, p)` or `S_of_axes(axes..., p)`
 where Matlab objects (in particular an `sqw` object) are passed as input to the user defined function,
-will not be possible with non-Matlab functions, unless `OptFunction` unwraps this object into a (several) POD array(s).
+will not be possible with non-Matlab functions, because the `a`, `sqw_obj` or `axes` objects cannot be passed from Matlab to an external library function.
+However, `OptFunction` may still accept such objects but will need rules for transforming them into POD coordinates (which it already understands for `sqw` or `dnd` objects).
 
 The next difficulty is to infer the required input arguments for user defined model functions.
 Several methods for this springs to mind:
@@ -112,6 +116,11 @@ This again will wrap the input arguments so that no memory copies occur, provide
 and the Python side function expects only `numpy` arrays as inputs.
 Again, like the first case, output arguments will have to be memory-copies
 or the user-defined function should fill in values in a Matlab-allocated `numpy` wrapped array.
+
+It should be noted here that the two use cases above will probably require *separate* implementations for the Matlab-Python interface.
+A compiled Horace library within Python (case 2) will crash if its Matlab component (e.g. Horace / `OptFunction` / `OptModel`)
+tries to use the Matlab-built-in Python interface, because this entails Matlab launching a separate Python process which conflicts with the existing Python process.
+
 
 
 ## <a name="compiled"></a> Compiled languages
@@ -589,8 +598,15 @@ Constructing the %-encoded link can be done at, e.g., the codecogs website
 	https://www.codecogs.com/eqnedit.php?latex=latex_equation
 --->
 [\mathbb{Q} \equiv (\mathb{Q},E)]:http://latex.codecogs.com/svg.latex?%5Cmathbb%7BQ%7D%5Cequiv%28%5Cmathbf%7BQ%7D%2CE%29
+[\mathbb{Q}]:http://latex.codecogs.com/svg.latex?%5Cmathbb%7BQ%7D
 [S(\mathbb{Q})]: http://latex.codecogs.com/svg.latex?S%28%5Cmathbb%7BQ%7D%29
 [\mathbf{Q}]: http://latex.codecogs.com/svg.latex?%5Cmathbf%7BQ%7D
 [\mathbb{Q}]: http://latex.codecogs.com/svg.latex?%5Cmathbb%7BQ%7D
 [N]: http://latex.codecogs.com/svg.latex?N
 [(E_n, S(\mathbf{Q}, E_n))]: http://latex.codecogs.com/svg.latex?%28E_n%2C%20S%28%5Cmathbf%7BQ%7D%2C%20E_n%29%29
+[(E_n, S(\mathbf{Q}, E_n), \gamma(\mathbf{Q}, E_n))]: http://latex.codecogs.com/svg.latex?%28E_n%2C%20S%28%5Cmathbf%7BQ%7D%2C%20E_n%29%2C%20%5Cgamma%28%5Cmathbf%7BQ%7D%2C%20E_n%29%29
+[\gamma(\mathbf{Q}, E_n)]: http://latex.codecogs.com/svg.latex?%5Cgamma%28%5Cmathbf%7BQ%7D%2C%20E_n%29
+[(|Q|, E)]: http://latex.codecogs.com/svg.latex?%28%7CQ%7C%2C%20E%29
+[S(|Q|, E)]: http://latex.codecogs.com/svg.latex?S%28%7CQ%7C%2C%20E%29
+[E]: http://latex.codecogs.com/svg.latex?E
+[S(E)]: http://latex.codecogs.com/svg.latex?S%28E%29
--- a/optimisation/design/Third_Party_API_Design.md
+++ b/optimisation/design/Third_Party_API_Design.md
@@ -31,7 +31,7 @@ the broadening function in `disp2sqw` will be a combination of both intrinsic an
 
 In the current Horace implementation, these model functions are used by the `multifit_sqw` function/class
 (which inherits from `multifit`) to fit to data using a built-in Levenberg-Marquardt optimizer. 
-The [new design for PACE](Model_Optimisation_Design.md) proposes instead to use a pair of classes, `OptFunction` and `OptModel`.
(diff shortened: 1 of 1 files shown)


optimisation/design/Third_Party_API_Design.md
```diff
@@ -0,0 +1,595 @@
...
+and because, as noted [in the appendix](app_nativememory) the main Matlab code and library codes have different memory heaps,
+so they must each be responsible for allocating *and deallocating* their own memory.
+That is, a memory error will occur if Matlab tries to deallocate a memory block allocated by the library and vice versa.
+To avoid this, *we propose to not allow user functions to allocate their own memory for the results array*.
+That is, Matlab will allocate an array for the results and pass the pointer to this to the user model function.
+The user model function must then fill this array with the result of the calculations,
+rather than using their own allocated array.
+
+Finally we note that the compiled library interface deal *exclusively* with real-valued variables.
+The user defined model function is supplied with real input coordinates ![\mathbb{Q}] and is expected to return real ![S(\mathbb{Q})].
```
nickbattam-tessella, line 176: `![\mathbb{Q}]` is being rendered raw

optimisation/design/Third_Party_API_Design.md
```diff
@@ -0,0 +1,595 @@
...
+
+```matlab
+function user_model_sqw(qh, qk, ql, en, parameters, varargin)
+function user_model_dsp(qh, qk, ql, parameters, varargin)
+function user_model_pow(modQ, en, parameters, varargin)
+function user_model_1d(en, parameters, varargin)
+function user_model_0d(parameters, varargin)
+```
+
+This follows the convention established by the current implementation of `multifit`.
```
nickbattam-tessella, line 193: Is it easier to pass 3 or 4 `1xn` arrays of values for `qh`, `qk`, `ql`, `en` or a single `3xn` or `4xn` array?

nickbattam-tessella (commented):
How much demand is there for the FORTRAN API? It appears this is the the "hardest" of the bindings to manage and implement.

optimisation/design/Third_Party_API_Design.md
mducle: For Matlab and Python user functions, probably it doesn't matter either way. For the compiled library there might be an ambiguity on the data layout - is the Matlab 2D array passed column or row-major? Newer Matlab supports C-style row-major matrices but the ones I've tested all return a column-major array.

Also, unless we add another `int` to the function signature (in addition to `n_elem`) to determine the size of the input array, we would need to trust that the users will handle the input array size correctly. With multiple 1D input arrays we can use the runtime signatures checking in `calllib` to ensure that the correct library function is called.

So, on balance I would prefer to stick with multiple 1D arrays, unless it would be much more performant for Horace to pass a 2-D array.

mducle:
> How much demand is there for the FORTRAN API? It appears this is the the "hardest" of the bindings to manage and implement.

I don't know how much demand there would be... I've only ever had one user actually wanting to use a compiled code with Horace and this code was written in Fortran 90. In the end we compiled it as an executable which reads in the Q points from a file which Horace wrote - not really optimal since it was a text file. Results were read back from another file. I also tried to use the built-in `perl` interpreter to run the process and use pipes for communication but I don't think I ever got that to work.

In terms of "hardest" - as far as Horace is concerned whether the user code is in C, C++ or Fortran is immaterial because it all looks like plain C to `loadlibrary` and `calllib`. The difficulties are mainly on the user side. We could make it *a bit* easier for the users but this would probably involve writing a separate Fortran `mex` file. However, I think that this would not be cross-compatible between compilers (I presume that `gfortran` and `ifort` etc. have different ABI). The advantage of using the C-style interface is that it uses the very basics of how libraries interface each other so there is the minimum of ABI-incompatibility issues. We'll probably always have linker issues and such depending on the users systems / compilers but that's not really some that can be avoided - we can just provide clear documentations and work-arounds.

optimisation/design/Third_Party_API_Design.md
```diff
@@ -0,0 +1,595 @@
...
+
+
+# Motivation
+
+Analysis of inelastic neutron scattering data almost always requires a parameterised model of the measured scattering intensities.
+Horace, as currently implemented, requires this model to be in the form of a Matlab function.
+This function is required to either 
+
+1. Take a 4D vector of momentum-energy ![\mathbb{Q} \equiv (\mathb{Q},E)] and yield the scattering function ![S(\mathbb{Q})].
+2. Take the 3D vector of momentum transfer ![\mathbf{Q}] only and yield ![N] sets of vectors ![(E_n, S(\mathbf{Q}, E_n))] where ![N] is the number of dispersion branches.
```
g5t, line 17: I would propose a third variant which takes **Q** and returns E<sub>n</sub>, S(**Q**,E<sub>n</sub>), and γ(**Q**,E<sub>n</sub>); where γ(**Q**,E<sub>n</sub>) *is* the intrinsic energy width of the n<sup>th</sup> mode at **Q**.

optimisation/design/Third_Party_API_Design.md
```diff
@@ -0,0 +1,595 @@
...
+Thus, the new PACE implementation should allow both types of models.
+We also note, that when instrumental resolution convolution (`TobyFit`) is used,
+the broadening function in `disp2sqw` should only represent some intrinsic physical broadening
+(such as a quasiparticle lifetime).
+Whereas when `TobyFit` is not used,
+the broadening function in `disp2sqw` will be a combination of both intrinsic and instrumental resolution broadening.
+
+In the current Horace implementation, these model functions are used by the `multifit_sqw` function/class
+(which inherits from `multifit`) to fit to data using a built-in Levenberg-Marquardt optimizer. 
+The [new design for PACE](Model_Optimisation_Design.md) proposes instead to use a pair of classes, `OptFunction` and `OptModel`.
```
g5t:
```suggestion
The [new design for PACE](Model_Optimisation_Design.md) proposes instead to use a trio of classes, `OptFunction`, `OptFunctions`, and `OptModel`.
```
or since `OptFunctions` probably isn't relevant for this discussion 
```suggestion
The [new design for PACE](Model_Optimisation_Design.md) proposes instead to use a host of classes including `OptFunction` and `OptModel`.
```

optimisation/design/Third_Party_API_Design.md
```diff
@@ -0,0 +1,595 @@
...
+`OptModel` will be more like the current `multifit` (or `tobyfit`) class,
+whilst `OptFunction` is a wrapper around the user supplied model function to handle parameter bindings and bounds currently handled by `multifit`.
+`OptFunction` should therefore have functionality to transform type 2 input to type 1 input
+whilst `OptModel` would expect only type 1 input.
```
g5t: I agree that `OptFunction` must contain the ability to transform function **output** into the equivalent of type 1.
In the case of resolution convolution the class inheriting from `OptFunction` and implementing the Tobyfit method would do this conversion at the start of its scope, but other convolution methods might perform the energy-axis convolution analytically in which case type 2 (or my suggested type 3) function output would be converted to type 1 as part of the convolution.

optimisation/design/Third_Party_API_Design.md
```diff
@@ -0,0 +1,595 @@
...
+
+# <a href="external_interfaces"></a> Matlab interfaces to external languages
+
+One immediate restriction on non-Matlab user-defined model functions is that its *input* arguments must be "plain-old-data" (POD, numeric types or strings) arrays only, and not objects.
+This is because Matlab does not allow its own objects (instances of Matlab-defined classes) to be passed to external libraries.
+(A workaround for this for the Python user interface using a compiled Horace library exists,
+and is described [here](../../python_interface/design/01_pace_python_high_level_discussion.md#interpreter-calls)).
+This means that the second type of function signatures described in the [optimisation design document](Model_Optimisation_Design.md#optimisation-function-object), 
+where Matlab objects (in particular an `sqw` object) are passed as input to the user defined function,
+will not be possible with non-Matlab functions, unless `OptFunction` unwraps this object into a (several) POD array(s).
```
g5t: `OptFunction` was designed with exactly this in mind. Wrapped functions will need to have their expected input specified (either explicitly or via some sort of function inspection) and `OptFunction` will contain the mechanism required to produce POD arrays when necessary.

optimisation/design/Third_Party_API_Design.md
```diff
@@ -0,0 +1,595 @@
...
+
+## <a name="python"></a> Python
+
+There are two use cases for user defined model functions written in Python:
+
+1. The user is running Matlab (has a license) and wants to call a model function (e.g. Euphonic) written in Python.
+2. The user is running Horace in Python using the compiled library and wants to call their own Python model function.
+
+Taking the first case, it is currently possible to do this in Matlab
+(as exemplified by the [`meuphonic` project](https://github.com/pace-neutrons/Meuphonic)).
```
g5t, line 97: `meuphonic` has been renamed [`horace-euphonic-interface`](https://github.com/pace-neutrons/horace-euphonic-interface), however the old link does and will forever continue to work.

optimisation/design/Third_Party_API_Design.md
```diff
@@ -0,0 +1,595 @@
...
+This requires, however, conversions between Matlab and Python data types,
+(as demonstrated by the `m2p` and `p2m` methods of [Brille](https://github.com/g5t/brille/)),
+which implicitly requires a memory copy in each direction
+(one memory copy of each input and output arrays for a single function evaluation).
+
+We can avoid the memory copy for the input argument arrays as long as they are restricted to be numeric arrays (`numpy` on Python-side) only.
+This would use a `mex` file to wrap the Python function call and its arguments in a `numpy` array,
+similar to the work described [here](../../python_interface/design/02_pace_python_implementation_discussion.md#pymatpy).
+
+As noted in the [Appendix](#app_nativememory), a memory copy of the output argument is only possible
```
g5t:
```suggestion
As noted in the [Appendix](#app_nativememory), avoiding a memory copy of the output argument is only possible
```
Is this missing a word?

optimisation/design/Third_Party_API_Design.md
```diff
@@ -0,0 +1,595 @@
...
+	[latex_equation]: %-encoded link to image generator
+then, if image generation fails for some reason, the alt-text for each equation
+will still contain the raw LaTeX equation which is often intelligible
+
+Constructing the %-encoded link can be done at, e.g., the codecogs website
+	https://www.codecogs.com/eqnedit.php?latex=latex_equation
+--->
+[\mathbb{Q} \equiv (\mathb{Q},E)]:http://latex.codecogs.com/svg.latex?%5Cmathbb%7BQ%7D%5Cequiv%28%5Cmathbf%7BQ%7D%2CE%29
+[S(\mathbb{Q})]: http://latex.codecogs.com/svg.latex?S%28%5Cmathbb%7BQ%7D%29
+[\mathbf{Q}]: http://latex.codecogs.com/svg.latex?%5Cmathbf%7BQ%7D
```
g5t, line 603:
```suggestion
[\mathbf{Q}]: http://latex.codecogs.com/svg.latex?%5Cmathbf%7BQ%7D
[\mathbb{Q}]: http://latex.codecogs.com/svg.latex?%5Cmathbb%7BQ%7D
```
To address the issue pointed out by Nick above.

optimisation/design/Third_Party_API_Design.md
```diff
@@ -0,0 +1,595 @@
...
+
+## <a name="matlab_functions"></a> Matlab
+
+```matlab
+function user_model_sqw(qh, qk, ql, en, parameters, varargin)
+function user_model_dsp(qh, qk, ql, parameters, varargin)
+function user_model_pow(modQ, en, parameters, varargin)
+function user_model_1d(en, parameters, varargin)
+function user_model_0d(parameters, varargin)
+```
```
g5t, line 191: What about the cases outlined in `OptFunction` where, e.g., plotting axes are passed to the function?
I propose that all prototypical function signatures be of the form `function user_model_NdX`? where `N` is how many equal-sized 1-D arrays are expected before the parameters and `X` indicates the return type listed above.

optimisation/design/Third_Party_API_Design.md
```diff
@@ -0,0 +1,595 @@
...
+
+As we are using `loadlibrary` / `calllib` to get Matlab to evaluate a function in a compiled library,
+we need to use C-calling conventions for these functions for all languages: C, C++ and Fortran.
+In order to accommodate Fortran, these functions *should pass by reference only*, and have no return values.
+
+The function signatures in C is thus:
+
+```c
+void user_model_sqw(const double *qh, const double *qk, const double *ql, const double *en,
+                    const double *parameters, double *results, int *n_elem);
```
g5t:
```suggestion
                    const double *parameters, double *results, unsigned long *n_elem);
```
`n_elem` might be *very* large, it should be at least an unsigned 32-bit integer (2<sup>32</sup>-1 ~ 4.29 billion).

Also, similar to above, we should define signatures for the N<=4 dimensional user functions of type 1.

optimisation/design/Third_Party_API_Design.md
```diff
@@ -0,0 +1,595 @@
...
+void user_model_dsp(const double *qh, const double *qk, const double *ql,
+                    const double *parameters, double *result_omega, double *results_S, int *n_elem);
+```
+
+where `n_elem` is the number of elements which Matlab expects `results` to have.
```
g5t:
```suggestion
where `n_elem` is the number of elements which Matlab expects `results` to have and the length of all input vectors.
```

optimisation/design/Third_Party_API_Design.md
mducle: Sorry I was not clear. I meant that the user-defined function signature: 
> `arrays_objects(a, ..., p)` of which `S_of_Q_w(sqw_obj, p)` is a specific case

Will not be possible for non-Matlab defined `array_objects` or `S_of_Q_w` because `a` or `sqw` cannot be passed from Matlab to an external library.

optimisation/design/Third_Party_API_Design.md
mducle: I think the users can name their functions whatever they like... the names are just to make explicit what physical type of data the mappings would correspond to... the users would have to specify which case is which in the `OptFunction` constructor below.

I'm assuming that `X` would always be a double.

optimisation/design/Third_Party_API_Design.md
mducle: This text is about the current implementation...

optimisation/design/Third_Party_API_Design.md
mducle: Ok. I've changed the text to note that for standard transformations listed in the optimisation document that `OptFunction` will automatically do the transformation of both input and output, and for others, the user is responsible for the transformation.

PR #11, opened 2020-07-01
Add ADRs for PACE project documentation decisions

nickbattam-tessella:
PACE-project level ADRs recording decisions reached with regard to documentation

### DIFF
--- a/documentation/adr/0002-use-github-pages-for-user-documentation.md
+++ b/documentation/adr/0002-use-github-pages-for-user-documentation.md
@@ -29,7 +29,8 @@ We will use GitHub pages for all PACE project documentation.
 
 - documentation will be available at URLs under `pace-neutrons.github.io` 
 - GitHub pages can be configured to make the documentation available at any owned URL
-- documentation is stored "as code" alongside the source
+- sources for the documentation are stored alongside the source code
+- built documentation is stored in the GitHub
 - build pipelines for each project will be required to convert the source `rst` files to `html` for display
 - API documentation can be included for compiled C code
 
--- a/documentation/adr/0002-use-github-pages-for-user-documentation.md
+++ b/documentation/adr/0002-use-github-pages-for-user-documentation.md
@@ -1,0 +1,35 @@
+[<-previous](0001-record-architecture-decisions.md) | [next->](0003-pace-projects-will-have-independent-documentation.md)
+
+# 2. Use GitHub Pages for user documentation
+
+Date: 2020-Jun-20
+
+## Status
+
+Accepted
+
+## Context
+
+A consistent platform is required for user documentation from the PACE projects: Brille, Euphonic and Horace/Herbert.
+
+Two platforms support, are well used for this service:
+
+- [Read the Docs](https://readthedocs.org/)
+- [GitHub pages](https://pages.github.com/)
+
+Both platforms will display documentation built by [Sphinx](https://www.sphinx-doc.org/) from reStructuredText source files.
+
+Brille includes compiled C libraries that contribute APIs to the project. The Read the Docs build/deploy pipeline does not support inclusion of documentation generated from C source.
+
+## Decision
+
+We will use GitHub pages for all PACE project documentation.
+
+## Consequences
+
+- documentation will be available at URLs under `pace-neutrons.github.io` 
+- GitHub pages can be configured to make the documentation available at any owned URL
+- documentation is stored "as code" alongside the source
+- build pipelines for each project will be required to convert the source `rst` files to `html` for display
+- API documentation can be included for compiled C code
+
--- a/documentation/adr/0005-pace-projects-must-be-semantically-versioned.md
+++ b/documentation/adr/0005-pace-projects-must-be-semantically-versioned.md
@@ -1,0 +1,31 @@
+[<-previous](0004-pace-documentation-link-to-projects.md) | next->
+
+# 5. PACE Projects must be semantically versioned
+
+Date: 2020-Jun-20
+
+## Status
+
+Accepted
+
+## Context
+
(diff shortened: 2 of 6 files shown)


documentation/adr/0002-use-github-pages-for-user-documentation.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Decision
+
+We will use GitHub pages for all PACE project documentation.
+
+## Consequences
+
+- documentation will be available at URLs under `pace-neutrons.github.io` 
+- GitHub pages can be configured to make the documentation available at any owned URL
+- documentation is stored "as code" alongside the source
```
rebeccafair: Does 'as code' mean the .rst or html?

documentation/adr/0005-pace-projects-must-be-semantically-versioned.md
```diff
@@ -0,0 +1,31 @@
...
+- `minor` version when functionality is added in a backwards compatible manner, and
+- `patch` version for backwards compatible bug fixes.
+
+Sphinx plugins are available that support multiple documentation versions ([sphinx-multiversion](https://pypi.org/project/sphinx-multiversion/)) - this will not work correctly for Brille where documentation is extracted from the build artifacts.
+
+## Decision
+
+Projects will be semantically versioned and documentation will be retained and be available for users to access for each major or minor version released. 
+
+There is no requirement to have patch-release specific documentation.
```
g5t, line 26: The GitHub [Action](https://github.com/g5t/brille/tree/master/.github/actions/build_sphinx) which builds the `brille` documentation (triggered [by the workflow here](https://github.com/g5t/brille/blob/master/.github/workflows/build_docs.yml)) stores _every_ versioned release separately in a folder in the `gh-pages` branch, e.g., [v0.3.0](https://g5t.github.io/brille/v0.3.0/).

Other than repository size, is there an argument against keeping documentation for patch-level releases?

nickbattam-tessella: My reasoning for that was that there should never been a need for documentation changes for a patch-level updates as these should not be adding features or changing APIs.

If the documentation is updated as part of a patch (fixing or clarifying an example) for the `m.n` release that update should apply equally to `m.n.0` and `m.n.198`

documentation/adr/0002-use-github-pages-for-user-documentation.md
nickbattam-tessella: Yes.

nickbattam-tessella: There are two cases here:

- The document source (i.e. `rst` and static `html` is part of the 'codebase`. 
- The built artefacts (`html`, `css`, ...) will be stored in the same GitHub repository, either in the `gh_pages` branch or a configured "built documentation" folder. 

That should be clarified

rebeccafair: Gotcha, it might be helpful to specify that 'code' means rst and static html. And I guess the rest will be documented in another ADR once we've decided exactly where to store the versioned build artefacts.

nickbattam-tessella: I should also an an extra ADR documenting the preference to use the `gh_pages` branch for built documentation and the consequences for branch and build management that creates.

documentation/adr/0006-store-built-documentation-in-branch.md
```diff
@@ -0,0 +1,63 @@
...
+
+- offers clear separation between source and built documentation
+- carries an overhead or managing another repository
+- documentation will served at a URL that doesn't match the source: e.g. `pace-neutrons.github.io/horace-docs`
+
+
+
+## Decision
+
+Built documentation will be stored on the `gh_pages` branch.
```
g5t:
```suggestion
Built documentation will be stored on the `gh-pages` branch.
```

documentation/adr/0006-store-built-documentation-in-branch.md
```diff
@@ -0,0 +1,63 @@
...
+
+
+
+## Consequences
+
+- The build pipeline will need to move the built pages from source branch (`master`) to the publication branch  (`gh-pages`)
+- Any previously build documentation for the current project version must be deleted from the `gh-pages` branch.
```
g5t, line 60:
```
- The build pipeline will need to move the built pages from source branch (`master`) to the publication branch  (`gh-pages`)
```
`sphinx-build` can direct its output to an out-of-source temporary directory, c.f., the [Action](https://github.com/g5t/brille/blob/2340fbf5b44d246275069f97c307191cb8e99b22/.github/actions/build_sphinx/entrypoint.sh#L49) in `brille`.

```
- Any previously build documentation for the current project version must be deleted from the `gh-pages` branch.
```
I suggest `rsync --delete`  from the out-of-source output directory to a **subfolder** in the `gh-pages` branch, c.f., the [Action](https://github.com/g5t/brille/blob/2340fbf5b44d246275069f97c307191cb8e99b22/.github/actions/build_sphinx/entrypoint.sh#L67) in `brille` -- this is a bit simpler than the `git rm`, `git reset`, `git checkout`, `rsync` method employed by. e.g, the Azure [Pipeline](https://github.com/scipp/scipp/blob/master/.azure-pipelines/templates/documentation_deploy.yml#L39-L77) in `scipp`.
As long as the subfolder name is related to the release/tag then updated documentation will replace the version which it supersedes. E.g., if the documentation for a release goes into a folder 'v**M**.**m**' (**M**ajor, **m**inor) then documentation for a patch release will overwrite the existing 'v**M**.**m**' documentation.

tgperring (approved):

PR #12, opened 2020-07-15
Developer script storage

rebeccafair:


mducle (approved):
Looks good to me.

documentation/adr/0007-developer-scripts-storage-location.md
```diff
@@ -0,0 +1,53 @@
...
+script is tied to a particular software, or general algorithm development
+it could be stored in `pace-developers/euphonic/scripts` or
+`pace-developers/powder_averaging/scripts` for example.
+
+
+## Consequences
+
+* Scripts that are intended for development only should not be committed to
+the master branch of main software repositories
+* Any development scripts committed to `pace-developers` should specify commit
```
nickbattam-tessella: The commit hash is less relevant than API version (e.g. the m.n release number for the project). 

Including the hash adds some additional information, particularly where these scripts are in active development against rapidly developing, unreleased APIs, but for legacy utilisation the version number is more transparent:
- a developer can quickly see a script written against the v1.3 API is unlikely to work without effort in the current master v3.4.n code
- the project documentation describing the API is versioned with the `m.n` project releases
- a git commit hash (e.g. `4e0510a`) gives the developer no information about compatibility or age and would require a search back through git history to find that information

rebeccafair: That's a fair point. We could have both, similar to the way [versioneer](https://github.com/warner/python-versioneer) does it e.g. `v0.11+g1076c97` which specifies both the most recent released version and the commit? (Versioneer actually also includes the number of commits since the last release too, but for projects that don't use versioneer that's not so easy to figure out, and we probably don't need that level of information).

ajjackson: Seems that the hash is more useful immediately and the semver is more useful in the long run. I'd be happy to use the combined notation if people don't find it too cumbersome?

ajjackson:
That looks perfect, allowing the hash to be dropped if you know it isn't needed.

PR #13, opened 2020-08-10
Script(s) for Euphonic powder-averaging development

ajjackson:
This is derived from scripts used to present initial findings.

As discussed in ADR0007, the script is located in a Euphonic-specific
directory, with a version tag at the top generated by versioneer.

Note that this tag, generated with `versioneer.get_version()`, does not exactly follow the format VERSION+HASH, it adds an additional number section to make VERSION+NN.HASH. Presumably this is something like an incremental number of commits? Is that a problem to include?

### DIFF
--- a/euphonic/design/06_powder_convergence_discussion.md
+++ b/euphonic/design/06_powder_convergence_discussion.md
@@ -150,6 +150,7 @@ Keith Refson, David Voneshen
 ## High-priority
    - Profile powder-averaging script to identify sources of
      poor performance (AJJ & RF)
+   - Investigate possibility of limiting memory usage by sampling in batches of q-points. (AJJ & RF)
    - Examine impact of jitter in r-direction; does this clear up outliers? (AJJ)
    - Implement user API and scripts to make use of proposed 3-parameter powder-averaging scheme.
      - Try min/max instead of geometric mean
@@ -161,4 +162,4 @@ Keith Refson, David Voneshen
      (Euphonic-Brille team)
    - Look into fully volumetric sampling (AJJ)
    - Consider interactions between powder-averaging and instrumental
-     resolution functions (?)
+     resolution functions in overall design (TP)
--- a/euphonic/design/06_powder_convergence_discussion.md
+++ b/euphonic/design/06_powder_convergence_discussion.md
@@ -13,7 +13,7 @@ Keith Refson, David Voneshen
         grid-based, random and quasirandom sampling schemes.
 
      - Optional parameters include the number of q-point samples,
-       Gaussian/Lorentian broadening width, and addition of
+       Gaussian/Lorentzian broadening width, and addition of
        jitter in phi/theta directions
 
    - Currently no tool is provided to apply such average over a sweep
@@ -45,7 +45,7 @@ Keith Refson, David Voneshen
      - Regular grids alias and are inefficient
      - Quasirandom "golden" sampling cannot be systematically subsampled
 
-   - TP suggests investigation of Sobel sampling method, which allows
+   - TP suggests investigation of Sobol sampling method, which allows
      progressive quasirandom sampling with blue noise-like distribution.
 
 ## Sampling parameter sensitivity tests
@@ -115,11 +115,16 @@ Keith Refson, David Voneshen
        agreement that such assumptions should be theoretically
        justified and backed-up empirically.
 
-     - Another cost-saver would be the use of Brille to map qpts to BZ
-       and interpolate the spectrum.
-       - This requires _periodic_ contributions to spectrum to be
-         separated from terms that depend on absolute _q_.
-       - Hopefully that is just the Debye-Waller term and it will work.
+     - Another cost-saver would be the use of Brille for Brillouin zone (BZ)
+        interpolation.
+       - In this case, Euphonic would be used to calculate the phonon eigenvalues
+         and eigenvectors on a grid in the first (or first irreducible) BZ.
+       - An arbitrary qpt is transformed by the symmetry operations of the space group
+         into the equivalent point in the first BZ, and the phonon eigenvalue/vector
+         at the qpt can be obtained by linear interpolation in the first BZ.
+       - The structure factor is calculated as normal from the eigenvalues/vectors
+       - There is the overhead of first computing the grid in the first BZ, and the
+         optimum grid size or point density should also be tested for convergence.
 
 # Incidental observations
   - The calculations were quite slow to run (even with
@@ -150,7 +155,7 @@ Keith Refson, David Voneshen
      - Try min/max instead of geometric mean
 
 ## Longer-term
-   - Look into Sobel sampling (AJJ)
+   - Look into Sobol sampling (AJJ)
    - Consider powder-averaging at large q as a good
      application/demonstration of Brille-Euphonic interface
      (Euphonic-Brille team)
--- a/euphonic/design/06_powder_convergence_discussion.md
+++ b/euphonic/design/06_powder_convergence_discussion.md
@@ -1,0 +1,159 @@
+# Powder averaging: exploration of convergence properties
+
+These notes were drawn up by Adam Jackson following a meeting on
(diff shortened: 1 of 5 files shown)


rebeccafair:
> Note that this tag, generated with `versioneer.get_version()`, does not exactly follow the format VERSION+HASH, it adds an additional number section to make VERSION+NN.HASH. Presumably this is something like an incremental number of commits? Is that a problem to include?

Yes it is the number of commits, and is quite a useful rough measure of 'change since release' so should probably be included! The reason I didn't specify it in the ADR is that for other projects that don't use versioneer, counting the number of commits since the last release would be impractical... I'll update the adr to explain this.

euphonic/design/06_powder_convergence_discussion.md
```diff
@@ -0,0 +1,159 @@
...
+## Background/methodology
+   - We intend to provide user-friendly orientational-averaging tools to
+     users, primarily for the computation of coherent S(|q|,ω) maps of
+     powdered systems.
+
+     - Spherical averaging functions are now included in Euphonic, with
+        grid-based, random and quasirandom sampling schemes.
+
+     - Optional parameters include the number of q-point samples,
+       Gaussian/Lorentian broadening width, and addition of
```
mducle:
```suggestion
       Gaussian/Lorentzian broadening width, and addition of
```

euphonic/design/06_powder_convergence_discussion.md
```diff
@@ -0,0 +1,159 @@
...
+     terminates when spectrum appears converged.
+   - Concern about stability with respect to _small_ changes in number
+     of samples, and cost effectiveness with respect to _large_ steps
+     in number of samples.
+   - Subsampling/progressive sampling does not play well with the
+     sampling schemes implemented so far in Euphonic
+     - Regular grids alias and are inefficient
+     - Quasirandom "golden" sampling cannot be systematically subsampled
+
+   - TP suggests investigation of Sobel sampling method, which allows
```
mducle:
```suggestion
   - TP suggests investigation of Sobol sampling method, which allows
```
https://en.wikipedia.org/wiki/Sobol_sequence

euphonic/design/06_powder_convergence_discussion.md
```diff
@@ -0,0 +1,159 @@
...
+     - DV raised that this may not be consistent with the generally
+       rigorous approach taken to accuracy in this project; general
+       agreement that such assumptions should be theoretically
+       justified and backed-up empirically.
+
+     - Another cost-saver would be the use of Brille to map qpts to BZ
+       and interpolate the spectrum.
+       - This requires _periodic_ contributions to spectrum to be
+         separated from terms that depend on absolute _q_.
+       - Hopefully that is just the Debye-Waller term and it will work.
```
mducle:
```suggestion
     - Another cost-saver would be the use of Brille for Brillouin zone (BZ)
        interpolation.
       - In this case, Euphonic would be used to calculate the phonon eigenvalues
         and eigenvectors on a grid in the first (or first irreducible) BZ.
       - An arbitrary qpt is transformed by the symmetry operations of the space group
         into the equivalent point in the first BZ, and the phonon eigenvalue/vector
         at the qpt can be obtained by linear interpolation in the first BZ.
       - The structure factor is calculated as normal from the eigenvalues/vectors
       - There is the overhead of first computing the grid in the first BZ, and the
         optimum grid size or point density should also be tested for convergence.
```

euphonic/design/06_powder_convergence_discussion.md
```diff
@@ -0,0 +1,159 @@
...
+
+## High-priority
+   - Profile powder-averaging script to identify sources of
+     poor performance (AJJ & RF)
+   - Examine impact of jitter in r-direction; does this clear up outliers? (AJJ)
+   - Implement user API and scripts to make use of proposed 3-parameter powder-averaging scheme.
+     - Try min/max instead of geometric mean
+
+## Longer-term
+   - Look into Sobel sampling (AJJ)
```
mducle:
```suggestion
   - Look into Sobol sampling (AJJ)
```

euphonic/design/06_powder_convergence_discussion.md
```diff
@@ -0,0 +1,159 @@
...
+   - Consider powder-averaging at large q as a good
+     application/demonstration of Brille-Euphonic interface
+     (Euphonic-Brille team)
+   - Look into fully volumetric sampling (AJJ)
+   - Consider interactions between powder-averaging and instrumental
+     resolution functions (?)
```
mducle: I'd sic Toby on this one!

euphonic/scripts/compare_spectra.py
```diff
@@ -0,0 +1,72 @@
...
+        fractional:
+            Calculate relative difference by formula (S - R)/R. If False,
+            use the unscaled difference between values instead.
+        threshold:
+            Ignore error from values smaller than this threshold in reference
+            spectrum when fractional=True
+
+    Returns:
+        array Quantity:
+            Average difference between spectra
```
mducle:
```suggestion
            Difference between spectra
```

mducle (approved):
Looks good. Thanks for this! 

I noticed there's a lot of duplicated code - e.g. the argument parser, `get_spectrum()`, `get_ref_spectrum()` and `plot()` functions - maybe this could be hived off into another module like `diff_1d` and `diff_1d_avg`?

euphonic/design/06_powder_convergence_discussion.md
```diff
@@ -0,0 +1,159 @@
...
+  - A basic implementation of numerical powder-averaging should be
+    included in the upcoming research paper and corresponding Euphonic
+    release.
+  - Investigation of smarter approaches and Brille integration will need to wait.
+
+# Actions
+
+## High-priority
+   - Profile powder-averaging script to identify sources of
+     poor performance (AJJ & RF)
```
rebeccafair, line 152: Maybe also add a note to 'investigate strategies to reduce memory usage' or similar

rebeccafair (approved):

euphonic/design/06_powder_convergence_discussion.md
ajjackson: done!

euphonic/design/06_powder_convergence_discussion.md
ajjackson: Done! (In these notes.) I have also added a note to https://github.com/pace-neutrons/Pace-Project-Plan/issues/116 which isn't assigned to anyone.

ajjackson:
> Looks good. Thanks for this!
> 
> I noticed there's a lot of duplicated code - e.g. the argument parser, `get_spectrum()`, `get_ref_spectrum()` and `plot()` functions - maybe this could be hived off into another module like `diff_1d` and `diff_1d_avg`?

Yes, there is scope for more tidying. But really I wonder if it is even a good idea to use the diff module; as 'scripts' these should be standalone.

mducle:
> But really I wonder if it is even a good idea to use the diff module; as 'scripts' these should be standalone.

Fair enough... I think it's 6 of one, half a dozen of another... 

On the one hand, in a separate module the code is marginally cleaner and I think we probably don't want users to muck with the difference algorithms.

On the other, if they really do want to calculate a different difference (say they want to compute the difference of the squares of the intensity for some obscure reason...) this would be much easier for them if the difference functions are in the main script body. Not sure how frequent this use case is though.

Another reason to include the difference functions in the main script rather than the module is so users can see clearly what they are calculating (e.g. what exactly the `threshold` and `fractional` options really mean in maths terms - e.g. they might be confused about whether the fractional option divides by the first input or reference spectrum or their root product or mean or whatever).

ajjackson:
The nicest place for the diff function to be would be as a `__sub__` method on the spectrum objects in Euphonic :smile: 

Not clear if the utility justifies the effort/maintenance though

mducle:
> The nicest place for the diff function to be would be as a `__sub__` method on the spectrum objects in Euphonic :smile:
> Not clear if the utility justifies the effort/maintenance though

Agreed! (on both counts)... might be best to leave as it is now (as a separate module)...

PR #14, opened 2020-09-02
Add initial Euphonic-Brille design document

rebeccafair:
There are several options for how the interface between Euphonic-Brille could be implemented, this document aims to discuss the potential options and advantages/disadvantages from Euphonic's point of view, and suggest example syntax (addresses [#133](https://github.com/pace-neutrons/Pace-Project-Plan/issues/133)).

Currently these are just initial thoughts that will hopefully initiate some discussion, and are completely up for criticism. It loosely references SpinW and partially addesses [#139](https://github.com/pace-neutrons/Pace-Project-Plan/issues/139), but should probably be expanded to consider wider PACE (although at that point this document should probably be kept somewhere else as it's no longer Euphonic design).

A while ago I thought it would be better to have Brille calling Euphonic, but while writing this document I feel perhaps it may actually be better the other way round, and would reduce the amount of duplicated code and number of different libraries required (brilleu, brillem, horace-euphonic-interface...). That may change again as we discuss things.

### DIFF
--- a/euphonic/design/06_brille_integration.md
+++ b/euphonic/design/06_brille_integration.md
@@ -23,7 +23,7 @@ Horace users if they were somewhat similar.
 ### General Note on Syntax
 
 There are two options for the syntax, one is to  try to keep the commands as
-specific as possible so each command has fewer arguments and its clearer what
+specific as possible so each command has fewer arguments and it is clearer what
 they do. This has the benefit of reducing the chance of name clashes, making
 arguments easier to manage and easier for the user to input. However, this will
 generally lead to the user having to write more code as they will need one line
@@ -52,7 +52,7 @@ current Horace simulation functions the syntax splits arguments up into 'model
 parameters' for fitting and 'extra parameters' for constant arguments. For now
 I've made an example with the scaling factor as a fitting parameter.
 
-```
+```matlab
 brille_init_args = {'grid', 'mesh'
                     'max_points', 10000,
                     'parallel', true};
@@ -125,7 +125,7 @@ doesn't need to find it itself).
 
 From Python:
 
-```
+```python
 from euphonic import ureg, ForceConstants
 from euphonic.util import mp_grid
 
@@ -152,7 +152,7 @@ functions to make it easier to simulate cuts, or whether it should more closely
 mirror Euphonic's Python interface. The following example is somewhere
 inbetween and assumes a Matlab wrapper has been written (named 'meuphonic').
 
-```
+```matlab
 % Create ForceConstants and BrilleForceConstants using Euphonic-like syntax
 fc = meuphonic.ForceConstants.from_castep('quartz.castep_bin');
 kwargs = {'use_c', true, 'n_threads', 4};
@@ -197,3 +197,14 @@ by Brille?
 **Disadvantages:**
 - Would it be worth adding a dependency on Brille for what could be a small
   gain?
+
+## Decision
+
+Largely motivated by the ability to use Brille to improve the performance of
+powder simulations from Python, it has been decided to call Brille from Euphonic.
+(As of 05/05/22) Brille has been specified as an optional dependency, so it
+would not affect other Euphonic users. Also, the `BrilleForceConstants` (now
+named `BrilleInterpolator`) can be instead created with
+`BrilleInterpolator.from_force_constants` syntax, so no changes to the current
+`ForceConstants` class are required. The relevant PR for Euphonic is
+[PR#104](https://github.com/pace-neutrons/Euphonic/pull/104).
--- a/euphonic/design/06_brille_integration.md
+++ b/euphonic/design/06_brille_integration.md
@@ -1,0 +1,199 @@
+# Brille-Euphonic Integration
+
+## Context
+
+[Brille](https://github.com/brille/brille) is a library for symmetry operations
+and linear interpolation within an irreducible part of the first Brillouin
+zone. In the context of Euphonic it means that it can be used to linearly
+interpolate phonon eigenvalues/vectors to calculate them more quickly. There
+are 2 ways in which this could work. One is where Brille calls Euphonic, which
+means it can be used for Horace cuts, and Euphonic doesn't get any added
+complexity, but also means Euphonic standalone users don't get any of the
+benefits. The other option is that Euphonic calls Brille, this would require
+changes to Euphonic but would allow faster interpolation for both Horace and
+Euphonic standalone users. However, these options are not necessarily exclusive
+and it's possible that both could be implemented.
+
+The purpose of this document is to assess the benefits and drawbacks of these
+options and define any desired syntax, which will hopefully reveal any
+potential issues. This document should also be considered in the context of
+what the Brille-SpinW interface will look like, as it would be beneficial to
+Horace users if they were somewhat similar.
+
+### General Note on Syntax
+
+There are two options for the syntax, one is to  try to keep the commands as
+specific as possible so each command has fewer arguments and its clearer what
+they do. This has the benefit of reducing the chance of name clashes, making
+arguments easier to manage and easier for the user to input. However, this will
+generally lead to the user having to write more code as they will need one line
+for reading force constants, one line for setting up Brille's grid, one line
+for computing the lookup table etc. and could get very verbose. The other
+option is to try to have as few calls as possible and try to encapsulate all
+the options in arguments to one or two functions. This would be easy for a user
+who is happy with default arguments, but could become very complicated as
+more arguments and features are added (e.g. powder averaging).
+
+## Calling Euphonic from Brille
+
+### Use Case 1: Simulating Horace cuts
+This would involve using Horace's simulation functions to call Brillem (which
+calls BrillEu, which calls Brille and Euphonic).
+
+**Example Syntax**
+
+This syntax tries to keep the Brille/Euphonic arguments deliberate and
+separate, however it could be confusing for some users (nested cell arrays)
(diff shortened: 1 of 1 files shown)


euphonic/design/06_brille_integration.md
```diff
@@ -0,0 +1,199 @@
...
+   matter?)
+
+### Use case 2: Brille Q-point Mesh for BZ Integration
+This would involve using Brille to produce a mesh of q-points and their weights
+which Euphonic could use for Brillouin Zone integration (instead of naively
+generating a Monkhorst-Pack grid using `mp_grid` as in previous examples, which
+makes no use of symmetry). This is a much 'lighter' use of Brille and should in
+theory be easy to implement. However, libraries that Euphonic already depends
+on such as `spglib` can already generate weighted q-point Monkhorst-Pack grids,
+so is there actually much to be gained from having a 'smarter' mesh generated
```
g5t, line 191:
```
so is there actually much to be gained from having a 'smarter' mesh generated
```
@krefson @ajjackson please correct me if I'm wrong, but one of the 'bad' things about MP grids is that they are always constructed from linear combinations of the basis vectors. So if the angle between two basis vectors is unusually small (or large) or the ratio between any two basis vector lengths is large a MP grid may be poorly conditioned for some calculations.
The 'grids' constructed by `brille` are always built in an orthonormal coordinate system as either a Delaunay tetrahedralisation or from a series of (truncated) cubic cells.  So the positions of the points in a `brille` grid (away from the irreducible Brillouin zone boundaries) do not depend on the basis vectors.

ajjackson: I think there's a fundamental question to be asked about the merits of linear interpolation for BZ integration. Is it actually going to improve the results? I suppose this is somewhat equivalent to the tetrahedron integration employed in some electronic structure calculations.

The work on generalised grids from Gus Hart's group shows improvement over naive Monkhorst-Pack sampling for electronic structure of realistic materials. https://arxiv.org/pdf/1804.04741.pdf I'd expect this to transfer to phonon sampling as well. I'm not sure how Brille's method compares; it's worth noting that these improved schemes are still based on linear combinations of basis vectors - but use better combinations!

An earlier piece of important work in this area is Moreno/Soler https://journals.aps.org/prb/abstract/10.1103/PhysRevB.45.13891 ; they illustrate the impact of different primitive cell choices. They also recommend optimising the grid origin for maximum symmetry-reduction. This part hasn't been popular; it's no fun to implement and I for one prefer to keep the GP in my calculations because it's usually interesting.

euphonic/design/06_brille_integration.md
```diff
@@ -0,0 +1,199 @@
...
+**Advantages:**
+ - Would benefit both Horace and Euphonic users
+ - Allows Horace users to skip a step depending on if they want to use brille
+   or not
+
+**Disadvantages:**
+ - Would require additions to Euphonic's ForceConstants, and quite a few
+   changes to Brille's existing repos before this option is feasible
+ - The interface could end up being quite different to SpinW (does this
+   matter?)
```
mducle, line 182: Just a comment on the current SpinW integration. For using SpinW through Brille, the syntax:
```matlab
tri = sw_model('triAF', 1); 
spinw_pars = {'hermit', false, 'formfact', true, 'resfun', 'gauss'}; 
brl_pars = {'max_volume', 1e-5};
brl = brillem.Brille(tri, brl_pars{:});
sim_cut = disp2sqw_eval(exp_cut, @brl.horace_sqw, spinw_pars, fwhm);
```
is currently useable. Brille specific options are passed to the constructor `brillem.Brille` and only SpinW related parameters are passed to `disp2sqw_eval`.

For using SpinW directly from Horace, the syntax is:
```matlab
tri = sw_model('triAF', 1); 
spinw_pars = {'hermit', false, 'formfact', true, 'resfun', 'gauss'}; 
sim_cut = disp2sqw_eval(exp_cut, @tri.horace_sqw, spinw_pars, fwhm);
```
E.g. the same as with Brille but with the two Brille related lines removed, and calling the `horace_sqw` method of the SpinW object directly. (`tri = sw_model('triAF', 1);` is just a convenient shortcut to create a simple spin wave model. Usually, users would need several lines to create a model themselves.)

The final use case is for users wanting to using Brille with SpinW without using Horace (calculating the linearly interpolated spin wave spectrum without reference to a Horace workspace). The syntax here is:
```matlab
tri = sw_model('triAF', 1); 
spinw_pars = {'hermit', false, 'formfact', true, 'resfun', 'gauss'}; 
brl_pars = {'max_volume', 1e-5};
brl = brillem.Brille(tri, brl_pars{:});
brl.fill(spinw_pars{:})
hkl_path = {[0 0 0] [1 1 1] [0 0 1] [0 0 0] 100}; 
spec = brl.spinwave(hkl_path);
sw_plotspec(sw_egrid(sw_neutron(spec), 'Evect', linspace(0,20,100)));
```
compared with the syntax without Brille:
```matlab
tri = sw_model('triAF', 1); 
spinw_pars = {'hermit', false, 'formfact', true, 'resfun', 'gauss'}; 
hkl_path = {[0 0 0] [1 1 1] [0 0 1] [0 0 0] 100}; 
spec = tri.spinwave(hkl_path, spinw_pars{:});
sw_plotspec(sw_egrid(sw_neutron(spec), 'Evect', linspace(0,20,100)));
```
The main difference is that the SpinW parameters has to be passed to the `fill` call which then passes all its arguments to the underlying calculator (e.g. SpinW). No parameters except the hkl path or grid should be passed to the Brillem `spinwave` method unlike for the pure-SpinW case where the SpinW parameters are passed there.

euphonic/design/06_brille_integration.md
```diff
@@ -0,0 +1,199 @@
...
+```
+
+From Matlab:
+
+The question here is whether the interface should be 'Horace-like' with bespoke
+functions to make it easier to simulate cuts, or whether it should more closely
+mirror Euphonic's Python interface. The following example is somewhere
+inbetween and assumes a Matlab wrapper has been written (named 'meuphonic').
+
+```
```
mducle:
```suggestion
```matlab
```

euphonic/design/06_brille_integration.md
```diff
@@ -0,0 +1,199 @@
...
+The purpose of this document is to assess the benefits and drawbacks of these
+options and define any desired syntax, which will hopefully reveal any
+potential issues. This document should also be considered in the context of
+what the Brille-SpinW interface will look like, as it would be beneficial to
+Horace users if they were somewhat similar.
+
+### General Note on Syntax
+
+There are two options for the syntax, one is to  try to keep the commands as
+specific as possible so each command has fewer arguments and its clearer what
```
mducle:
```suggestion
specific as possible so each command has fewer arguments and it is clearer what
```

euphonic/design/06_brille_integration.md
```diff
@@ -0,0 +1,199 @@
...
+separate, however it could be confusing for some users (nested cell arrays)
+and some settings you may want to apply for both Brille and Euphonic rather
+than having to set them separately (e.g. if you want to use 4 threads in
+Euphonic to create the lookup table, you probably also want to use 4 threads in
+Brille to linearly interpolate the eigenvalues/vectors). In addition, with the
+current Horace simulation functions the syntax splits arguments up into 'model
+parameters' for fitting and 'extra parameters' for constant arguments. For now
+I've made an example with the scaling factor as a fitting parameter.
+
+```
```
mducle:
```suggestion
```matlab
```

euphonic/design/06_brille_integration.md
```diff
@@ -0,0 +1,199 @@
...
+eigenvectors), although this might not be required if calculations are
+appropriately chunked. Another update to make this integration easier would be
+adding symmetry information to Euphonic's `Crystal` object (so that Brille
+doesn't need to find it itself).
+
+**Example Syntax**
+
+From Python:
+
+```
```
mducle:
```suggestion
```python
```

euphonic/design/06_brille_integration.md
mducle: The current Brille/SpinW syntax is very similar to the proposed Euphonic syntax so I think this is fine. The main difference between `brillem.Brille` (the standalone Brille Matlab interface) and `brillem.Euphonic` (the Brille+Euphonic Matlab interface) is that `brillem.Euphonic` is just a light wrapper for `brilleu` (a Python module interfacing Brille to Euphonic) and so does not have many methods to interact with the Brille object. This is pertinent in the separation of keyword arguments or input parameters meant for Brille or for the calculator (Euphonic or SpinW). In the case of `brillem.Euphonic` both sets should be passed to the constructor at the same time; For `brillem.Brille` the constructor does not immediately construct the grid (e.g. invoke the calculator) so it only needs Brille specific arguments. Only secondary calls to `fill` (the `horace_sqw` method implicitly calls `fill` if a grid is not constructed or is out-of-date) need to be passed the arguments for the calculator (SpinW in the above examples). This separation could be useful or annoying depending on your point of view... (it could also make it harder to implement universal arguments which should apply to both Brille and the calculator).

mducle (approved):
Just a few minor suggestions for changes and a comment on how Brille/SpinW are integrated (Matlab only).

cmarooney-stfc:
@rebeccafair Hi Becky - noting this is still open and over a year old. What should we be doing with it? Thanks Chris

rebeccafair:
Sorry, I completely forgot about this... I will resolve the comments and add a conclusion on where we are with the syntax (I don't think it will change much now) and merge this, hopefully will get to this later today

PR #15, opened 2020-09-22
Add proposed syntax integrating Brille-SpinW-Horace

mducle:
Adds a document proposing syntax which users could use when running a combination of Brille / SpinW / Horace and/or MSlice to analyse powder and single-crystal neutron ToF spectroscopy data.

### DIFF
--- a/brille/design/Proposed_integrated_syntax_Brill_SpinW_Horace.md
+++ b/brille/design/Proposed_integrated_syntax_Brill_SpinW_Horace.md
@@ -1,0 +1,267 @@
+# Proposed Integrated Syntax when using a combination of Horace, SpinW, Brill and MSlice
+**2020-09-22**
+
+## Table of Contents
+
+[Introduction](#introduction)
+
+[Workflow](#workflow)
+
+[Extracting the data](#extracting-the-data)
+
+[Modelling the data](#modelling-the-data)
+
+[Fitting the data with resolution convolution](#fitting-the-data-with-resolution-convolution)
+
+[Implications for implementations](#implications-for-implementations)
+
+## Introduction
+
+[Brille](https://github.com/brille/brille) is a C++ library with Python wrappings for symmetry operations and linear interpolation within an irreducible part of the first Brillouin Zone.
+It can be used to construct the first Brillouin zone, find the high symmetry points and the irreducible polyhedron of any spacegroup.
+In addition, it can be used to linearly interpolate any quantity (vector or scalar) which obeys the symmetries of the input spacegroup.
+It is examples of this interpolating functionality that we shall explore in this document, specifically focusing on applications to magnetic time-of-flight neutron spectroscopy data.
+
+This document will include both working (mostly Matlab) and proposed (mostly Python) code examples.
+Proposed code examples will be marked by black lines on either side of the code block.
+The code examples will use a combination of Brille, SpinW, Horace and MSlice.
+
+## Workflow
+
+The workflow is:
+
+1. [Horace](https://github.com/pace-neutrons/Horace/) or *MSlice* is first used to extract (a subset of) the data to be modelled or fitted.
+2. SpinW or Brille is then used to simulate the extracted data with some set of model parameters. 
+The returned results should be of the same class as the data such that they may be easily plotted by Horace or MSlice.
+3. If the model has parameters (e.g. SpinW), it should be possible to fit those parameters to the measured data.
+4. For both fitting and simulation of a model, it should be possible to incorporate the instrumental resolution in the calculcation.
+
+## Extracting the data
+
+For single crystal data, this is currently possible in Matlab using Horace:
+
+```matlab
+proj = projaxes([1 0 0], [0 1 0]);
+cut1 = cut_sqw(sqw_file, proj, [hmin hstep hmax], [kmin kstep kmax], [lmin lmax], [emin estep emax]);
+```
+
+The above example assumes that data processing to construct an `.sqw` file combining multiple measurements of a crystal at different orientations is done.
+Triplets of numbers indicate that the data is to be rebinned along that axis (defined by the projection `proj`), whilst doublets indicates the data should be integrated between those ranges.
+Thus the example generates a 3D `sqw` object `cut1` from the 4D data in the `sqw_file`.
+
+The proposed syntax for Python is:
+
+<table>
+<td><img src="https://placehold.it/5x100/000000/000000?text=+"></td>
+<td>
+<pre lang="python">
+from pyHorace import horace as h
+proj = {'u':[1,0,0], 'v':[0,1,0], 'type':'rrr'}
+cut1 = h.cut_sqw(sqw_file, proj, [hmin, hstep, hmax], [kmin, kstep, kmax], 
+                                 [lmin, lmax], [emin, estep, emax])
+</pre>
+</td>
+<td><img src="https://placehold.it/5x100/000000/000000?text=+"></td>
+</table>
+
+For powder data, this is currently possible using the MSlice program.
+There are two implementations of MSlice: [one](https://github.com/mantidproject/mslice) in Python as part of [Mantid](https://www.mantidproject.org)
+and [the original](http://mslice.isis.rl.ac.uk/Main_Page) written in Matlab.
+The original Matlab MSlice is mostly a GUI program with an undocumented CLI (by using the underlying functions of the GUI).
+This program is also currently not being maintained so we will not use it.
+
+In Python, the syntax to extract data from a powder measurement using Mantid MSlice is:
+
+```python
+import mslice.cli as m
+dat2 = m.Load(nxs_file)
+cut2 = m.Slice(d2, f'|Q|,{qmin},{qmax},{qstep}', f'DeltaE,{emin},{emax},{estep}')
+```
+
+Note that unlike in Horace, the original pixel information is not retained in this rebinning operation.
+
+We propose the following syntax for powder data handling in Horace, should it be implemented:
+
+<table>
+<td><img src="https://placehold.it/5x70/000000/000000?text=+"></td>
+<td>
+<pre lang="matlab">
+proj = projaxes('powder');
+cut2 = h.cut_sqw(sqw_file, proj, [qmin, qstep, qmax], [emin, estep, emax]);
+</pre>
+</td>
+<td><img src="https://placehold.it/5x70/000000/000000?text=+"></td>
+</table>
+
+This syntax should retain pixel information just like the equivalent single crystal invocation.
+
+
+## Modelling the data
+
+We assume that [SpinW](https://spinw.org/) would be used to model the magnetic excitations.
+To set up and evaluate a model on single crystal data with current syntax in Matlab is:
+
+```matlab
+tri = sw_model('triAF', 1); 
+spinw_pars = {'hermit', false, 'formfact', true, 'resfun', 'gauss'}; 
+sim1 = disp2sqw_eval(cut1, @tri.horace_sqw, spinw_pars, fwhm);
+```
+
+This same syntax should also be possible (unchanged) for powder data, which has not been implemented.
+
+In Python it would be:
+
+<table>
+<td><img src="https://placehold.it/5x70/000000/000000?text=+"></td>
+<td>
+<pre lang="python">
+tri = h.sw_model('triAF', 1)
+spinw_pars = {'hermit': False, 'formfact': True, 'resfun': 'gauss'}
+sim1 = h.disp2sqw_eval(cut1, tri.horace_sqw, fwhm, **spinw_pars)
+</pre>
+</td>
+<td><img src="https://placehold.it/5x70/000000/000000?text=+"></td>
+</table>
+
+[//]: # (**comment)
+
+In addition to using SpinW directly, it is possible to use Brille to interpolate a SpinW model (in Matlab):
+
+```matlab
+tri = sw_model('triAF', 1); 
+spinw_pars = {'hermit', false, 'formfact', true, 'resfun', 'gauss'}; 
+brl_pars = {'max_volume', 1e-5};
+brl = brillem.Brille(tri, brl_pars{:});
+sim_cut = disp2sqw_eval(exp_cut, @brl.horace_sqw, spinw_pars, fwhm);
+```
+
+And the proposed Python syntax:
+
+<table>
+<td><img src="https://placehold.it/5x150/000000/000000?text=+"></td>
+<td>
+<pre lang="python">
+import brille as b
+tri = h.sw_model('triAF', 1)
+spinw_pars = {'hermit': False, 'formfact': True, 'resfun': 'gauss'}
+brl_pars = {'max_volume':1e-5}
+brl = b.Brille(tri, **brl_pars)
+sim1 = h.disp2sqw_eval(cut1, brl.horace_sqw, fwhm, **spinw_pars)
+</pre>
+</td>
+<td><img src="https://placehold.it/5x150/000000/000000?text=+"></td>
+</table>
+
+This syntax assumes that the Horace python distribution (`pyHorace`) will include SpinW as well as Horace.
+This would enable both to use the same Matlab interpreter process, and make intercommunication much easier and faster.
+Again these syntax should work for powder data objects, possibly replacing `h.disp2sqw` with `m.disp2sqw` method of `MSlice` (yet to be implemented).
+
+
+## Fitting the data with resolution convolution
+
+Horace includes a fitting framework which uses a Levenberg-Marquardt minimizer.
+The following example uses Brille and the TobyFit Monte Carlo resolution convolution method:
+
+```matlab
+proj = projaxes([1 0 0], [0 1 0]);
+cut1 = cut_sqw(sqw_file, proj, [hmin hstep hmax], [kmin kstep kmax], [lmin lmax], [emin estep emax]);
+model = pcsmo(JF1, JA, JF2, JF3, Jperp, D);   % User model defined as a function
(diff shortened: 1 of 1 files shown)


brille/design/Proposed_integrated_syntax_Brill_SpinW_Horace.md
```diff
@@ -0,0 +1,267 @@
...
+The proposed syntax for Python is:
+
+<table>
+<td><img src="https://placehold.it/5x100/000000/000000?text=+"></td>
+<td>
+<pre lang="python">
+from pyHorace import horace as h
+proj = {'u':[1,0,0], 'v':[0,1,0], 'type':'rrr'}
+cut1 = h.cut_sqw(sqw_file, proj, [hmin, hstep, hmax], [kmin, kstep, kmax], 
+                                 [lmin, lmax], [emin, estep, emax])
```
ajjackson, line 67: This is, effectively, 13 positional arguments and seems error-prone. Is it possible for some of the groups to be named optional arguments with useful defaults?

Also, pedantically, input groups like these are generally tuples rather than lists. So combining those we get a function signature like:
```python
def cut_sqw(sqw_file: Path, proj: Dict[str, Any],
            h_range: Optional[Tuple] = None, k_range: Optional[Tuple] = None,
            l_limits: Optional[Tuple] = None, e_range: Optional[Tuple] = None) -> WhateverTypeACutHas
```
and a "typical" usage would be something like e.g. `cut1 = h.cut_sqw(sqw_file, proj, l_limits=(l_min, l_max))`. `None` values are interpreted to use the full available range and standard step sizes.

Is that feasible, or am I missing some context here?

brille/design/Proposed_integrated_syntax_Brill_SpinW_Horace.md
```diff
@@ -0,0 +1,267 @@
...
+model_pars = {'mat':['JF1', 'JA', 'JF2', 'JF3', 'Jperp', 'D(3,3)'],
+              'hermit':False, 'optmem', 0, 'useFast':True, 'formfact':True,
+              'resfun':'gauss', 'coordtrans':np.diag([2 2 1 1])}
+brl_pars = {'max_volume', 1e-5};
+brl = b.Brille(model, brl_pars{:});
+sample = h.IX_sample(True,[0,0,1],[0,1,0],'cuboid',[0.01,0.05,0.01]);
+maps = h.maps_instrument(ei, freq, 'S');
+cut1.set_sample(sample);
+cut1.set_instrument(maps);
+tbf = tobyfit(cut1);
```
ajjackson: As a general rule I'm not a fan of the design pattern:
```python
object = MyClass()
object.set(**essential_data)
object.do_the_thing()
```
where some elegant alternatives are `MyClass(**essential_data).do_the_thing()` or even `do_the_thing(**essential_data)`.
The problem is that these workflows often allow one to construct objects in a "useless" state. This state needs to be tested, methods should fail elegantly if the required `set()` has not been performed yet, and users need to be taught to follow a sequence of actions rather than satisfying the input arguments of a single action.

I'm not very familiar with this tool/workflow, so it may be that there is a good reason that tobyfit is instantiated before setting properties and then fitting is called separately. And it's not as offensive as my example because there is at least _something_ going into init! But as I see it is imitating the Matlab workflow, it seems worth reviewing! Perhaps at least some steps would make sense to bundle together? Which steps are useful to separate (e.g. to perform multiple fitting variations after a slow data-loading step?)

brille/design/Proposed_integrated_syntax_Brill_SpinW_Horace.md
```diff
@@ -0,0 +1,267 @@
...
+
+This same syntax should also be possible (unchanged) for powder data, which has not been implemented.
+
+In Python it would be:
+
+<table>
+<td><img src="https://placehold.it/5x70/000000/000000?text=+"></td>
+<td>
+<pre lang="python">
+tri = h.sw_model('triAF', 1)
```
rebeccafair, line 137: It seems strange to access SpinW via `h.sq_model` when SpinW isn't really part of Horace. From your comments I understand it's so that they can use the same Matlab process, so maybe at least it should be in its own namespace `h.spinw.sq_model`?

If we wanted to have parity we could also have `h.euphonic` which would call `meuphonic` (or however we decide to call Euphonic from Matlab). Although then we get into the weird situation where we have Python calling Matlab calling Python, and I don't know what that would do to performance or even if that would be possible. My gut tells me it might be a bad idea.

I didn't include calling Horace from Python in my design document so I'll probably have to think about this more.

brille/design/Proposed_integrated_syntax_Brill_SpinW_Horace.md
```diff
@@ -0,0 +1,267 @@
...
+
+## Modelling the data
+
+We assume that [SpinW](https://spinw.org/) would be used to model the magnetic excitations.
+To set up and evaluate a model on single crystal data with current syntax in Matlab is:
+
+```matlab
+tri = sw_model('triAF', 1); 
+spinw_pars = {'hermit', false, 'formfact', true, 'resfun', 'gauss'}; 
+sim1 = disp2sqw_eval(cut1, @tri.horace_sqw, spinw_pars, fwhm);
```
rebeccafair, line 131: I know this is probably out of scope and already well established, but looking at `@tri.horace_sqw`, it seems strange to me that SpinW should have a top-level function specifically for producing Horace Sqw's. I feel like it potentially should be the other way round, that Horace contains functions to go from the SpinW/Euphonic model to its own Horace sqw...

brille/design/Proposed_integrated_syntax_Brill_SpinW_Horace.md
```diff
@@ -0,0 +1,267 @@
...
+<table>
+<td><img src="https://placehold.it/5x150/000000/000000?text=+"></td>
+<td>
+<pre lang="python">
+import brille as b
+tri = h.sw_model('triAF', 1)
+spinw_pars = {'hermit': False, 'formfact': True, 'resfun': 'gauss'}
+brl_pars = {'max_volume':1e-5}
+brl = b.Brille(tri, **brl_pars)
+sim1 = h.disp2sqw_eval(cut1, brl.horace_sqw, fwhm, **spinw_pars)
```
rebeccafair: I guess this is to get around Python calling Matlab calling Python? Again, I feel like having a Horace-specific top-level function (`horace_sqw`) in a generic library like Brille is wrong. There must be a better pattern for this somehow...

Will pyHorace be just a thin wrapper around Horace functions or is there scope to do some additional handling within Python? For disp2sqw_eval calling Python user functions (or Brille/Euphonic/pySpinW) there could be a case for doing some extra handling to keep the code mainly in Python with some calls to Horace or Euphonic (but no chaining of Horace->Euphonic) to get around the Python-Matlab-Python issue. Although I probably really need to think about this more and understand how pyHorace works/will work...

rebeccafair (commented):
Sorry in advance, big questions but not many answers!

brille/design/Proposed_integrated_syntax_Brill_SpinW_Horace.md
mducle: @ajjackson The context is that it's the syntax used in Matlab which doesn't have key-word arguments. I agree it's an ugly syntax for Python, and your suggestion with default values would be better.

The range arguments are actually with respects to the `u`, `v`, `w` directions defined by the projection `proj` rather than `h`, `k`, `l` necessarily. It's only because I've defined in this case u||h, v||k, w||l that it's the case here (this projection could also be the default), giving:

```python
def cut_sqw(sqw_input: Union[Path, sqw],     # input is either path to sqw file or sqw object in memory
            proj: Dict[str, Any] = None,     # Defaults to: {'u':[1,0,0], 'v':[0,1,0], 'type':'rrr'}
            u_range: Optional[Tuple] = None,
            v_range: Optional[Tuple] = None,
            w_range: Optional[Tuple] = None,
            e_range: Optional[Tuple] = None) -> sqw    # output is sqw object in memory or file-backed
```

with the default values for the range to be from `-inf` to `+inf` in steps equal to that stored in the data.

brille/design/Proposed_integrated_syntax_Brill_SpinW_Horace.md
mducle: Yeah, this is a bit tricky... We (e.g. Toby) came to the decision that the `sample` and `instrument` should be properties of the data object rather than the fit object since it can be passed around multiple fit objects; however it's not essential to look at the data (only essential for modelling it) so it should not be included in the constructor for the `sqw` object `cut1` (e.g. in the `cut_sqw` function). 

However, it *is* essential for the resolution convolution modelling, so maybe it should be included in the `tobyfit` constructor instead of in the data object... In which case the above calls would look like:

```python
tbf = h.tobyfit(cut1,
                h.IX_sample(True,[0,0,1],[0,1,0],'cuboid',[0.01,0.05,0.01]);
                h.maps_instrument(ei, freq, 'S'))
```

brille/design/Proposed_integrated_syntax_Brill_SpinW_Horace.md
mducle: Yeah, this is a bit tricky. 

I guess if we split things into different namespaces we should do something like:

```python
from pyPACE import pySpinW as sw
from pyPACE import pyHorace as h
```

where the big compiled library ("PACE") has separate namespaces for each program, but they are in the same Matlab compiled library and share the same interpreter / engine.

Which leads to the second "gotcha" which is that Matlab forbids the exchange of data between different Matlab interpreters / engines spawn by Python. This is even in the case where they might be in the same process and share memory (it's because the two engines are loaded from different `dll` and the libraries might have different ABI so that the memory layout is not compatible between them.) We can override this with some mex-file hackery but we would also have to ensure that if we distribute separate libraries for Horace and SpinW that the two libraries should be built by the same compiler with the same compiler options.

I agree that importing `meuphonic` (or whatever it's eventually called) into Python would not be a good idea, but the `pyPACE` library (or maybe "distribution"?) could also have the *Python* version of euphonic included instead.

brille/design/Proposed_integrated_syntax_Brill_SpinW_Horace.md
mducle: It's an interesting philosophical point, actually: Should *Horace* provide the interface to "user written" model or just provide an API, and *is* SpinW (and Euphonic) a "user written" (i.e. third party) model? I guess I went with the "just provide an API" option and treated SpinW the same as another third party code - the fact that we're all contributing to these codes and they are interconnected just blurs the line a bit... I guess we could discuss this at the meeting.

In fact, Horace does have such [interface functions](https://github.com/pace-neutrons/Horace/blob/master/horace_core/functions/spinw_sqw.m). These functions were actually written first and if I hadn't had any involvement with SpinW it would probably be the way we'd still do it. The issue is that there are certain coordinate transformations and optimisations in the model evaluation steps which needed modifications to the SpinW object itself. We found that ([a](https://github.com/brille/brillem/blob/master/%2Bbrillem/%40Brille/spinwave.m) [bit](https://github.com/brille/brillem/blob/master/%2Bbrillem/%40Brille/neutron_spinwave_intensity.m) [like](https://github.com/brille/brillem/blob/master/%2Bbrillem/%40Brille/sw_sab.m) Brillem we had to repeat some code in Horace or to store Horace information in a blank field in the SpinW object to pass from iteration to iteration).

In the end, because I was a SpinW contributor, it was easier to deal with it in SpinW. This could be a SpinW specific thing though. So the basic usage of SpinW is to define the model with all the exchange parameters, geometry etc, and then to calculate the spin wave dispersion at some set of `hkl`. During fitting we don't want to do all the set up steps, just to change a (sub-)set of exchange parameters (or similar). It is this "performance optimisation" which required changes to the SpinW code. This wouldn't really apply to Euphonic since you're not going to be varying the force constants.

brille/design/Proposed_integrated_syntax_Brill_SpinW_Horace.md
mducle: There will be scope for additional data handling; and I do envisage `disp2sqw` being able to call Python use functions. We'll have to discuss how much we want to reimplement Horace functionalities in Python... for a while one end-goal of PACE was a Python implementation of Horace but this was changed to just having a Python interface to Matlab code because we figured it would take too much developer effort to port everything from Matlab to Python.

g5t: I agree that anything like `horace_sqw` should not be a method of the generic (C++/Python) `brille`.
I guess that this is where my lack of imagination in naming becomes a problem.

In the example MATLAB code above `@brl.horace_sqw` is a method of the `brillem.Brille` class, which is a MATLAB class to handle setting-up and interacting with a generic `brille` interpolation grid through the `brille` Python bindings.

The proposed Python interface would need an interface similar to (MATLAB) `brillem.Brille`. This new interface could be in a submodule of the `brille` Python interface but should probably be defined in a new Python module (not unlike `brilleu`).

brille/design/Proposed_integrated_syntax_Brill_SpinW_Horace.md
```diff
@@ -0,0 +1,275 @@
...
+The prototype implementation, whose syntax is described below does not have these separate submodules however.
+This may be implemented later to support more "Pythonic" syntax, which would require more extensive Python wrapper code.
+
+
+## Workflow
+
+The workflow is:
+
+1. [Horace](https://github.com/pace-neutrons/Horace/) or *MSlice* is first used to extract (a subset of) the data to be modelled or fitted.
+2. SpinW or Brille is then used to simulate the extracted data with some set of model parameters.
```
cmarooney-stfc, line 47: Can Brille "simulate"? doesn't sound quite right.

brille/design/Proposed_integrated_syntax_Brill_SpinW_Horace.md
```diff
@@ -0,0 +1,275 @@
...
+The Python wrapper module for this library will have submodules for Horace (`pyHorace.horace`), SpinW (`pyHorace.spinw`) and general Matlab functions (`pyHorace.matlab`).
+The prototype implementation, whose syntax is described below does not have these separate submodules however.
+This may be implemented later to support more "Pythonic" syntax, which would require more extensive Python wrapper code.
+
+
+## Workflow
+
+The workflow is:
+
+1. [Horace](https://github.com/pace-neutrons/Horace/) or *MSlice* is first used to extract (a subset of) the data to be modelled or fitted.
```
cmarooney-stfc, line 46: reference for Mslice? say you're going to reference it below?

brille/design/Proposed_integrated_syntax_Brill_SpinW_Horace.md
```diff
@@ -0,0 +1,275 @@
...
+This would enable both to use the same Matlab interpreter process, and make intercommunication much easier and faster.
+
+Again these syntax should work for powder data objects, possibly replacing `h.disp2sqw`
+with a (yet to be implemented) `m.disp2sqw` method of `MSlice`.
+
+
+## Fitting the data with resolution convolution
+
+Horace includes a fitting framework which uses a Levenberg-Marquardt minimizer.
+The following example uses Brille and the TobyFit Monte Carlo resolution convolution method:
```
cmarooney-stfc, line 172: Maybe add "although other fitting methods also use the framework" to clarify it isn't just Tobyfit

brille/design/Proposed_integrated_syntax_Brill_SpinW_Horace.md
```diff
@@ -0,0 +1,275 @@
...
+```matlab
+proj = projaxes([1 0 0], [0 1 0]);
+cut1 = cut_sqw(sqw_file, proj, [hmin hstep hmax], [kmin kstep kmax], [lmin lmax], [emin estep emax]);
+model = pcsmo(JF1, JA, JF2, JF3, Jperp, D);   % User SpinW model defined as a function
+model_pars = {'mat', {'JF1', 'JA', 'JF2', 'JF3', 'Jperp', 'D(3,3)'}, ...
+              'hermit', false, 'optmem', 0, 'useFast', true, 'formfact', true, ...
+              'resfun', 'gauss', 'coordtrans', diag([2 2 1 1]), ...
+              'use_brille', 'node_volume_fraction', 1e-5, 'Qtrans', diag(1./[4 4 1])};
+sample = IX_sample(true,[0,0,1],[0,1,0],'cuboid',[0.01,0.05,0.01]);
+maps = maps_instrument(ei, freq, 'S');
```
cmarooney-stfc, line 183: the lattice parameters alatt and angdeg have now been moved from the generic run data to the sample - as they are sample-specific data. Once the sample object has been generated at line 182 you need to add
> sample.alatt = value
> sample.angdeg = value

brille/design/Proposed_integrated_syntax_Brill_SpinW_Horace.md
```diff
@@ -0,0 +1,275 @@
...
+
+[Modelling the data](#modelling-the-data)
+
+[Fitting the data with resolution convolution](#fitting-the-data-with-resolution-convolution)
+
+[Implications for implementations](#implications-for-implementations)
+
+## Introduction
+
+[Brille](https://github.com/brille/brille) is a C++ library with Python wrappings for symmetry operations and linear interpolation within an irreducible part of the first Brillouin Zone.
```
cmarooney-stfc, line 20: I thought that Brille did symmetry and interpolation from the whole of q-space ONTO an irreducible part of the BZ

cmarooney-stfc (changes requested):
Again a few changes. Wondering if the discussion at the end has now moved on as the document is quite old

PR #17, opened 2020-10-19
ADR for PACE integration syntax guiding principles

mducle:
Add an architectural design record for the guiding principles for designs of UI and interfaces between PACE project codes (e.g. between Horace/Herbert and Euphonic/SpinW/Brille etc).

This was discussed and agreed at a meeting on Sept 28th [(minutes here)](https://stfc365.sharepoint.com/:w:/r/sites/PACEProject/Shared%20Documents/Meetings/PACE-General/pace_syntax_meeting1_minutes.docx) to discuss the proposed syntax for such interactions between the codes from a user's perspective.

(I'm not sure that this quite the correct place for this to go...)

### DIFF
--- a/documentation/adr/0008-pace-syntax-guiding-principles.md
+++ b/documentation/adr/0008-pace-syntax-guiding-principles.md
@@ -44,6 +44,9 @@ Specifically the same method and keyword names should be used.
 reasonable defaults for as many input arguments as practicable);
 This applies to both Matlab and Python interfaces,
 since the end goal is to migrate entirely to Python.
+* Since the Matlab and Python *languages*'s argument syntax are not a superset,
+exactly what (sub-set) of syntax PACE should use should be clearly defined
+in a separate architectural decisions record.
 * In the mean time, there should be both Matlab and Python interfaces for 
 all functionalities (including handling polycrystalline/powder data).
 * There should be uniform API for interactions between the different programs
--- a/documentation/adr/0008-pace-syntax-guiding-principles.md
+++ b/documentation/adr/0008-pace-syntax-guiding-principles.md
@@ -1,0 +1,44 @@
+[<-previous](0007-developer-scripts-storage-location.md) | next->
+
+# 8. PACE Syntax Guiding Principles
+
+Date: 2020-Oct-19
+
+## Status
+
+Proposed
+
+
+## Context
+
+Whilst the core framework of PACE is the Horace (and Herbert) projects,
+PACE involves many other codes, including [Euphonic](https://github.com/pace-neutrons/Euphonic),
+[Brille](https://github.com/brille/brille) and [SpinW](https://spinw.org).
+The way in which these programs interact with each other is presently unclear.
+Furthermore the programs are written in several different languages.
+In this ADR we propose some guiding principles in the design of the interfaces
+between the programs or components which make up PACE.
+
+## Decision
+
+The guiding principle for PACE-supported programs is **commonality**.
+That is, interacting with the different programs of PACE should be **seamless**
+to the user, and the user interface presented to them should be **uniform**.
+
+## Consequences
+
+There are several implications stemming from this guiding principle:
+
+* There should be a single downloadble distribution which bundles all the 
+necessary components (e.g. Horace, Euphonic, Brille and SpinW at the moment).
+* User syntax should follow the same format for each program,
+and should be similar between Matlab and Python interfaces.
+Specifically the same method and keyword names should be used.
+* Syntax should be "Pythonic"(prefer keyword to positional arguments;
+reasonable defaults for as many input arguments as practicable);
+This applies to both Matlab and Python interfaces,
+since the end goal is to migrate entirely to Python.
+* In the mean time, there should be both Matlab and Python interfaces for 
+all functionalities (including handling polycrystalline/powder data).
+* There should be uniform API for interactions between the different programs
+(and with external third party programs).
--- a/documentation/adr/0008-pace-syntax-guiding-principles.md
+++ b/documentation/adr/0008-pace-syntax-guiding-principles.md
@@ -18,6 +18,8 @@ The way in which these programs interact with each other is presently unclear.
 Furthermore the programs are written in several different languages.
 In this ADR we propose some guiding principles in the design of the interfaces
 between the programs or components which make up PACE.
+This was discussed at a meeting whose
+[minutes are here.](https://stfc365.sharepoint.com/:w:/r/sites/PACEProject/Shared%20Documents/Meetings/PACE-General/pace_syntax_meeting1_minutes.docx))
 
 ## Decision
 
@@ -25,16 +27,20 @@ The guiding principle for PACE-supported programs is **commonality**.
 That is, interacting with the different programs of PACE should be **seamless**
 to the user, and the user interface presented to them should be **uniform**.
 
+The uniform interface means that function names and how users specify parameters
+or arguments to functions in Matlab and Python flavours of PACE should align
+as closely as possible.
+
 ## Consequences
 
 There are several implications stemming from this guiding principle:
 
 * There should be a single downloadble distribution which bundles all the 
-necessary components (e.g. Horace, Euphonic, Brille and SpinW at the moment).
+necessary components (presently: Horace, Euphonic, Brille and SpinW).
 * User syntax should follow the same format for each program,
 and should be similar between Matlab and Python interfaces.
 Specifically the same method and keyword names should be used.
-* Syntax should be "Pythonic"(prefer keyword to positional arguments;
+* Syntax should be "Pythonic" (prefer keyword to positional arguments;
 reasonable defaults for as many input arguments as practicable);
 This applies to both Matlab and Python interfaces,
 since the end goal is to migrate entirely to Python.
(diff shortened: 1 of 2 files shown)


documentation/adr/0008-pace-syntax-guiding-principles.md
```diff
@@ -0,0 +1,44 @@
...
+The guiding principle for PACE-supported programs is **commonality**.
+That is, interacting with the different programs of PACE should be **seamless**
+to the user, and the user interface presented to them should be **uniform**.
+
+## Consequences
+
+There are several implications stemming from this guiding principle:
+
+* There should be a single downloadble distribution which bundles all the 
+necessary components (e.g. Horace, Euphonic, Brille and SpinW at the moment).
```
cmarooney-stfc: Excuse being picky. e.g. means "for example" but it sounds as if this list is currently exhaustive. So just "currently: Horace .... etc". Maybe say Horace/Herbert for completeness.

documentation/adr/0008-pace-syntax-guiding-principles.md
```diff
@@ -0,0 +1,44 @@
...
+* User syntax should follow the same format for each program,
+and should be similar between Matlab and Python interfaces.
+Specifically the same method and keyword names should be used.
+* Syntax should be "Pythonic"(prefer keyword to positional arguments;
+reasonable defaults for as many input arguments as practicable);
+This applies to both Matlab and Python interfaces,
```
cmarooney-stfc, line 45: Since the MATLAB inputParser is a bit flaky for optional string arguments before keyword arguments, this could be stronger; we should only want compulsory positional and keyword arguments.

documentation/adr/0008-pace-syntax-guiding-principles.md
```diff
@@ -0,0 +1,44 @@
...
+Furthermore the programs are written in several different languages.
+In this ADR we propose some guiding principles in the design of the interfaces
+between the programs or components which make up PACE.
+
+## Decision
+
+The guiding principle for PACE-supported programs is **commonality**.
+That is, interacting with the different programs of PACE should be **seamless**
+to the user, and the user interface presented to them should be **uniform**.
+
```
cmarooney-stfc, line 29: These are excellent goals. It would be useful to illustrate this so everyone knows what this means in practice. There is one suggestion on function parameters in the consequences; this sounds more extensive and maybe we should say if any others are on the plan.

documentation/adr/0008-pace-syntax-guiding-principles.md
```diff
@@ -0,0 +1,44 @@
...
+## Context
+
+Whilst the core framework of PACE is the Horace (and Herbert) projects,
+PACE involves many other codes, including [Euphonic](https://github.com/pace-neutrons/Euphonic),
+[Brille](https://github.com/brille/brille) and [SpinW](https://spinw.org).
+The way in which these programs interact with each other is presently unclear.
+Furthermore the programs are written in several different languages.
+In this ADR we propose some guiding principles in the design of the interfaces
+between the programs or components which make up PACE.
```
cmarooney-stfc, line 20: We discussed this in the Horace/PACE team, and it seemed useful to record (a) who is proposing (b) when they proposed (c) what record there was of this discussion. This would give better traceability of hwo this was arrived at.

documentation/adr/0008-pace-syntax-guiding-principles.md
g5t: Alternatively, replace `e.g.` by `i.e.,` which is 'that is.'

On the point of a single distribution: do we want to commit to packaging disparate Python and MATLAB projects into a single archive (e.g., `.tar.gz`, `.zip`, `.exe`, or the like) or would an installation _script_ be a better alternative?
At the moment `Euphonic` and `Brille` are easily installed using `pip` and I wonder if the something similar could be accomplished for `Horace/Herbert` and `spinW` through MATLAB Toolboxes.

mducle:
```suggestion
necessary components (currently: Horace, Euphonic, Brille and SpinW).
```

documentation/adr/0008-pace-syntax-guiding-principles.md
cmarooney-stfc: I've now caught up with the location of the minutes of the meeting (opening comment, above) and it would be good just to have that location recorded in the ADR.

Additional thought: is github the best place to put these documents? We have the sharepoint, which may be more appropriate.

mducle: I've put the minutes on sharepoint and changed the links. Github keeps the file attachment forever though so anyone with the old link can still access it.

documentation/adr/0008-pace-syntax-guiding-principles.md
mducle: I think we would have to have separate Python and Matlab packages.

The problem is bundling everything together:
* For the Matlab distribution, how do we install the Python packages?
  - It basically depends on what distribution of Python the user wants to use. The `pyenv` and `pyversion` commands can be set up to use any install Python. The easiest might be to have a script to ask the user which Pythons they want and then to install the Python dependencies using `pip` to that Python distribution. This could accommodate people wanting to use Mantid's Python, Anaconda or the system Python.
  - The alternative is that we bundle our own Python distribution in an executable installer (for Windows) or create a package dependency on the system Python and then use that (Linux packagers).
* For the Python distribution, how to install the Matlab component?
  - This is a bit more straightforward since it definitely requires the Matlab MCR. The choice here is between bundling it or asking users to download it themselves. Bundling it means our build artifacts will be bigger by ~2GB. And I'm not so sure about the licensing for this. If you use their deployment tool, Matlab allows you to bundle the MCR but I'm not sure they allow you to do it in your own installers...
   - The alternative is to tell users they have to install the MCR themselves.
   - The third way is to have the installation script download the MCR for users and start running it so they just have to click-through the license.

Two other alternatives are to use containers (e.g. `docker`) or cross-platform packaging system (e.g. `conda`).

I sense another meeting about this...

documentation/adr/0008-pace-syntax-guiding-principles.md
mducle: Do you mean that _all_ positional and keyword arguments would _always_ have to be specified? This would severely restrict the flexibility of the syntax. I would definitely not want to impose this on the Python side. 

What about if we ban non-keyword string arguments? (e.g. the current `'-nopix'` would change to `'pix', false`)

(Or if non-keyword string arguments always have to be prefixed with `'-'` which would still allow `'-nopix'`)

cmarooney-stfc: Sorry, didn't mean that. At the moment the MATLAB input parser offers compulsory positional, optional positional and keyword. Roughly corresponding to Python compulsory positional, *args and **kwargs. Keyword arguments are always optional. I am suggesting not having any *args, to keep things simple. If we had a way of validating optional positional string arguments to distinguish them from the first keyword, that could be possible (as you say, e.g. -nopix), but maybe for simplicity have all arguments keyword?

mducle: Ah right. Sorry. That sounds fine then. We can omit the optional positional arguments `*args`.

cmarooney-stfc: Actually, realising that I'm getting my Python syntax wrong. But it does seem that Python and MATLAB have very different ways of processing optional arguments, (and also keyword arguments: MATLAB has them as a separate category while Python has them as an alternative way of accessing the same compulsory argument set - do I have that right?) - so a minimal set of ways that work for both would maybe be the way forward. Anyway, we can discuss at the next meeting.

mducle: So I think the compulsary/optional and positional/keyword arguments are not really part of the Matlab language as such, they're only part of the `inputParser` function. The Matlab language only really defines `varargin` which is a cell array and is equivalent to Python's `*args`. Matlab simulates Python's `**kwargs` using string-value pairs in `varargin`, so then it could have problems with string position arguments. Python, on the other hand has both positional and keyword arguments as part of the language itself. 

You're right that in Python you can access a named positional argument using a keyword with the same name as you give the positional argument variable. E.g. `def fun(a, b, c)` can be called with `fun(1, 2, 3)` _or_ `fun(a=1, b=2, c=3)` or `fun(1, 2, c=3)` with the convention that positional arguments must come first (so `fun(a=1, 2, 3)` will give a `SyntaxError`). Matlab does not support this in `inputParser` but I think if we write our own parser we could implement this if we wanted as long as we pre-define the names of the position arguments for the parser rather than just use the names in the function definition.

So, as I see it we can do one of two things in Matlab:

1. In the function definition we accept only `varargin` - that is, `function [x,y,z] = somefun(varargin)`. This cell array is passed to a parser which contains a list of *named* arguments *in order* with flags on whether they are keywords or positional, e.g. something like `[struct('name', 'arg1', 'kind', 'positional', 'type', 'numeric', 'default', 0), struct('name', 'arg2', 'kind', 'keyword', 'type', 'numeric', 'default', 0)]`. This way the user can call either `somefun('arg1', 1, 'arg2', 2)` *or* `somefun(1, 'arg2', 2)` like in Python. In this case I think we could allow the use of optional positional arguments unless the positional argument is exactly the same as a keyword argument name (we'd follow the Python convention that after the first keyword argument, every other argument is a keyword argument). 

2. Use a mix of defined (mandatory) positional arguments and `varargin` - that is `function [x,y,z] = somefun(arg1, arg2, varargin)`. In this case `varargin` contains *only* keyword arguments (its length must be even), and we disallow optional positional arguments. So the user would have to use `somefun(1, 2, 'arg3', 3)`.

PR #18, opened 2020-11-17
[Refactor] Remove generated math image links

g5t:
- Links to an external public server which generates images of math equations were not always properly cached by GitHub due to, probably, too-long response times from the generating server. The cached images were then displayed as broken and could not easily be fixed.
- All remote generated math images are now pre-generated and stored within this repository in a svg/ directory relative to each Markdown file. A Python utility is added to a new tools/ directory which can parse a Markdown file for codecogs.com svg image links, download those images and save them to a new svg/ directory, and update the Markdown links to point at the local files.
- With these changes the GitHub Markdown renderer no longer needs to cache the remote images and broken images should no longer occur.


mducle (approved):
It all looks fine. The only issue I can see is possible filename collisions. One immediate example I can see is `X_j` and `X^j` yields the same `Xj.svg` file. If both `X_j` and `X^j` appear the same folder then you get a `Missing URL?` error message for one of them. At the moment, the user can just go in and change the link name e.g. from `[X_j]:` to `[X_j2]:` but with the same codecogs url so it now generates `Xj2.svg`. 

This is probably ok for now, since creating the links etc (and running the `replace_codecogs_links` script) is a pretty manual process anyway.

If it becomes a problem, a solution might be to append the actual codecogs link as a comment to the end of the generate `svg` which the python script can then re-read to check for name collisions... or in the block which writes the `svg` (around line 30) to generate an `md5sum` or similar for the generated file which can then be stored in a dict and rechecked in an `else` block in that code...

g5t:
Thanks for the feedback @mducle. I also noticed that the regular expression used to identify inline image links, e.g., `link name` was too greedy and would also capture following text if there was a subsequent closing parentheses on the same line.
I've fixed that bug and updated the `replace_codecogs_links.py` script to use a slightly less likely to collide link name sanitiser, and to write the link name and url into a comment in each SVG file.

Running the updated script on the three Markdown files in `master/optimisation` and `master/resolution` finished without finding any name collisions, where there was one previously.
Commit 88219bb corrects the greedy-regex and name collision problems.

PR #20, opened 2020-11-29
Recorded ADR for brille/X vs X/brille

mducle:
Record decision reached at a meeting in October for calculators `X` to call `brille` for interpolation rather than reverse, and for these integration code to be in the calculator `X` rather than in `brille` or a third project.

### DIFF
--- a/documentation/adr/0009-brille-integration.md
+++ b/documentation/adr/0009-brille-integration.md
@@ -1,0 +1,41 @@
+[<-previous](0008-pace-syntax-guiding-principles.md) | next->
+
+# 9. Integration of Brille within PACE projects
+
+Date: 2020-Nov-29
+
+## Status
+
+Accepted
+
+
+## Context
+
+[brille](https://github.com/brille/brille) is a library for computing symmetry operations
+and linear interpolation within an irreducible part of the first Brillouin zone.
+Whilst its symmetry operations functionality can be used stand-alone,
+the interpolation functionality should be integrated with codes
+(such as [euphonic](https://github.com/pace-neutrons/euphonic) and [spinW](https://github.com/spinw/spinw))
+which compute quantities in reciprocal space to be user friendly.
+
+At present there are separate projects, [brilleu](https://github.com/brille/brilleu/)
+and [brillem](https://github.com/brille/brillem/) to achieve this integration.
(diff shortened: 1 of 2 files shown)


documentation/adr/0009-brille-integration.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Context
+
+[brille](https://github.com/brille/brille) is a library for computing symmetry operations
+and linear interpolation within an irreducible part of the first Brillouin zone.
+Whilst its symmetry operations functionality can be used stand-alone,
+the interpolation functionality should be integrated with codes
+(such as [euphonic](https://github.com/pace-neutrons/euphonic) and [spinW](https://github.com/spinw/spinw))
+which compute quantities in reciprocal space to be user friendly.
+
```
cmarooney-stfc, line 21: Only comment here is that the text says that euphonic and spinw compute quantities in reciprocal space to make those quantities user friendly. I think you mean that the interpolation functionality should be integrated with the codes to make them or it more user friendly.

rebeccafair (approved):
Looks good to me!

documentation/adr/0009-brille-integration.md
```diff
@@ -0,0 +1,42 @@
...
+should not be done by the external projects `brilleu` and `brillem`.
+Instead, the interface between the calculator X and brille should be embedded within X which will
+construct is own brille object and then use this to perform the interpolation.
+
+## Consequences
+
+* Code in `brilleu` will either be merged into `euphonic` or `brilleu` will be a dependency of `euphonic`
+* This will allow `euphonic` to use `brille` to compute the density of states, the Debye-Waller factor or powder averaging.
+* Code specific to `spinW` in `brillem` will be moved to `spinW` itself,
+leaving a minimal `brillem` for Matlab users to access the symmetry operations functionality of `brille`.
```
g5t:
```suggestion
leaving a minimal `brillem` for Matlab users to access the functionality of `brille`.
```
I thought we agreed that `brillem` should still contain a bare-bones interface to the `brille` 'grids' and their linear interpolation routines.

g5t (changes requested):
Other than not wanting to limit the scope of `brillem` this all looks good.

PR #22, opened 2020-12-02
ADR decision to use mex for Python <-> compiled Matlab calls

mducle:
Record decision to use the `mex` interface to allow compiled Matlab to call Python functions in `pyHorace`.

### DIFF
--- a/documentation/adr/0010-use-mex-for-pyHorace-python-calls.md
+++ b/documentation/adr/0010-use-mex-for-pyHorace-python-calls.md
@@ -1,0 +1,56 @@
+[<-previous](0009-brille-integration.md) | next->
+
+# 10. pyHorace Matlab code should use a mex interface to call Python functions
+
+Date: 2020-Dec-02
+
+## Status
+
+Proposed
+
+
+## Context
+
+`pyHorace` ([prototype](https://github.com/mducle/hugo)) is the proposed Python module for PACE
+which packages the Matlab Horace/Herbert codes as a "compiled" library.
+The Matlab code within this library needs to interact with Python code in the host.
+As detailed in [this design document](../../01_pace_python_high_level_discussion.md#-calling-user-defined-functions)
+there are two main methods for the compiled Matlab code to call a Python function:
+
+1. Using a `mex` file.
+2. Using a "reverse communications interface" (RCI).
+
+In the `mex` method, a reference to the Python function is given to the `mex` file which uses the CPython
+API to call that function (in Python) and then automatically converts (copies) the results to Matlab (`Array`) format.
+In the RCI approach, the standard Matlab-Python interface is used but is restricted to specific use of `multifit`.
+This is because the standard Matlab-Python interface does not allow the compiled Matlab code to call Python functions
+directly, it only allows the Python interpreter to call the compiled Matlab functions.
+Thus, to allow `multifit` to work, the Matlab `multifit` class needs to be refactored to allow it to _receive_ the residuals
+rather than to call a function evaluate a model and calculate the residuals itself.
+In addition, a Python wrapper class is needed which Python users call instead of the Matlab `multifit`.
+This Python wrapper can then call the user model, calculate the residual and send it (at each iteration) to the Matlab `multifit`.
+
+Thus, whilst the RCI uses the documented API (which should make it more stable and robust) 
+it is very restrictive and requires significant changes to `multifit` (which will happen in any case). 
+The `mex` option allows the current Matlab code to be used without modifications and is more flexible,
+but can result in hard crashes (segmentation faults).
+In addition, `mex` functions _cannot_ use non-Matlab (non-`Array` types) data as input or outputs.
+Thus all data must be converted to/from Matlab from/to Python, which involves at least one data copy.
+This restriction means that not all Python functions can be supported by this method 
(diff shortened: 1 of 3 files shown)


documentation/adr/0010-use-mex-for-pyHorace-python-calls.md
```diff
@@ -0,0 +1,56 @@
...
+Thus all data must be converted to/from Matlab from/to Python, which involves at least one data copy.
+This restriction means that not all Python functions can be supported by this method 
+(e.g. those which return non-convertible types such as Python classes).
+This also means that the reference to the Python function cannot be passed directly to the `mex` function.
+Rather, the Python code stores it in a global dictionary and the (string) key is passed to the `mex`.
+
+
+## Decision
+
+The decision was taken to use the `mex` method for compiled Matlab code to call Python functions.
```
cmarooney-stfc: There appear to be a number of good points for each of the two proposed solutions. So it is surprising not to see reasons why the decision has come down on this side. I think recording the reasons would be good.

documentation/adr/0010-use-mex-for-pyHorace-python-calls.md
```diff
@@ -0,0 +1,56 @@
...
+The Matlab code within this library needs to interact with Python code in the host.
+As detailed in [this design document](../../01_pace_python_high_level_discussion.md#-calling-user-defined-functions)
+there are two main methods for the compiled Matlab code to call a Python function:
+
+1. Using a `mex` file.
+2. Using a "reverse communications interface" (RCI).
+
+In the `mex` method, a reference to the Python function is given to the `mex` file which uses the CPython
+API to call that function (in Python) and then automatically converts (copies) the results to Matlab (`Array`) format.
+In the RCI approach, the standard Matlab-Python interface is used but is restricted to specific use of `multifit`.
```
cmarooney-stfc: You mean the MATLAB multifit function calls python rather than multifit being called? The wording makes the direction of the call a bit ambiguous; it would be good to be more explicit.

documentation/adr/0010-use-mex-for-pyHorace-python-calls.md
```diff
@@ -0,0 +1,56 @@
...
+This is because the standard Matlab-Python interface does not allow the compiled Matlab code to call Python functions
+directly, it only allows the Python interpreter to call the compiled Matlab functions.
+Thus, to allow `multifit` to work, the Matlab `multifit` class needs to be refactored to allow it to _receive_ the residuals
```
cmarooney-stfc: that is, receive the residuals from the python code which it is calling. Again it may be good to be a bit more explicit in the wording.

documentation/adr/0010-use-mex-for-pyHorace-python-calls.md
```diff
@@ -0,0 +1,56 @@
...
+rather than to call a function evaluate a model and calculate the residuals itself.
+In addition, a Python wrapper class is needed which Python users call instead of the Matlab `multifit`.
+This Python wrapper can then call the user model, calculate the residual and send it (at each iteration) to the Matlab `multifit`.
```
cmarooney-stfc: and now I am confused because the python code is now calling multifit (or at least would like to, but it is calling the wrapper instead.) At this point I think a diagram showing the calling sequence could be good, Undoubtedly the design document referenced will tell me more, but it is good to ensure that this can be read stand-alone if possible.

cmarooney-stfc (changes requested):
Various comments

mducle:
@cmarooney-stfc Sorry I think I wasn't clear on how the RCI works. It's because I wanted to keep this ADR brief - but I guess it is too brief.

Essentially, with the standard Matlab-Python interface for a compiled Matlab program, the compiled Matlab program *cannot* call any function defined by Python. It's set up so that only Python can call functions defined by the compiled Matlab program. What we want is a way around this restriction. 

So, first we need a way for the compiled Matlab program to let Python know that it wants to call a Python function. Then we need data exchange. The RCI approach is *only* about data exchange. That is, in the *specific* case of `multifit` we write a Python class (say `multifit_py`) which the user calls and passes their model function reference. This class then calls the Matlab `multifit` and tells it to do all the set up and everything up to the point where it would have tried calling the user function, and at this point returns control to `multifit_py` returning the values of the parameters at which the model should be evaluated. `multifit_py` then calls the user's model function with these parameters, obtains the result and then calls Matlab `multifit` again saying: "here is the data you would have gotten had you been able to call the user's function". `multifit` then computes the residuals, Jacobian etc and determines the next set of parameters and returns that to `multifit_py`, etc.

That's what I meant about this approach being specific `multifit`. If we want a general way for the compiled Matlab program to call a Python function we need some way of the compiled Matlab letting Python knows it wants to call a Python function. The simplest way of doing this is to use a `mex` file with CPython bindings. The other way I can think of is to create a server/client interface where we have a Python thread to listen for Matlab requests - but this just seems very complicated to me.

----

On your other points about reasons for the decision - essentially it's only been me making it; and I thought the RCI is too restrictive and needed too much changes to multifit so I went ahead with the `mex` prototype. However, it would be good to have a full discussion about it (hence why this ADR is "proposed").

g5t (approved):

PR #23, opened 2021-01-01
ADR to use light python wrapper

mducle:
Adds an architectural design record to propose to use a wrapper around Python code for use in Matlab for the Matlab interfaces for Brille and Euphonic.

### DIFF
--- a/documentation/adr/0012-matlab-python-wrapper.md
+++ b/documentation/adr/0012-matlab-python-wrapper.md
@@ -1,4 +1,4 @@
-[<-previous](0011-use-mex-for-pyHorace-python-calls.md) | [next->]()
+[<-previous](0011-use-mex-for-pyHorace-python-calls.md) | [next->](0013-light-python-wrapper-implementation-detail.md)
 
 # 12. A Matlab wrapper for Python PACE projects
 
@@ -6,18 +6,18 @@ Date: 2021-Jan-01
 
 ## Status
 
-Proposed
+Accepted
 
 
 ## Context
 
-Both Euphonic and Brille are PACE projects with primarily a Python interface.
-PACE, however, aims to provide both a Python and a Matlab interface to users,
+Both Euphonic and Brille are PACE projects with primarily a Python user interface (UI).
+PACE, however, aims to provide both a Python and a Matlab UI to users,
 and also to foster inter-operability between projects which are written both in Matlab and Python.
 In particular, `pyHorace` ([prototype](https://github.com/mducle/hugo)) cannot use the 
 [standard method](https://uk.mathworks.com/help/matlab/call-python-libraries.html) for Matlab to run Python code, 
 where calls to Python from Matlab are prefixed with `py.` followed by the full module specification.
-For example, `r = py.numpy.random.rand(3)` uses `numpy` to generate a random number.
+For example, `r = py.numpy.random.rand()` uses `numpy` to generate a random number.
 This is because such a call causes Matlab to 
 [automatically spawn](https://uk.mathworks.com/help/matlab/ref/pyenv.html) a dependent Python interpreter,
 which can be either created within the same process as the Matlab interpreter (`InProcess`)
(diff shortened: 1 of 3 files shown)


documentation/adr/0012-matlab-python-wrapper.md
```diff
@@ -0,0 +1,53 @@
...
+Proposed
+
+
+## Context
+
+Both Euphonic and Brille are PACE projects with primarily a Python interface.
+PACE, however, aims to provide both a Python and a Matlab interface to users,
+and also to foster inter-operability between projects which are written both in Matlab and Python.
+In particular, `pyHorace` ([prototype](https://github.com/mducle/hugo)), cannot use the standard
+method for Matlab to run Python code, using the `py.*` namespace.
```
rebeccafair: It's worth stating more clearly what the [standard method is](https://www.mathworks.com/help/matlab/call-python-libraries.html) and why `pyHorace` can't use it. It's alluded to below with the mention of crashes but could do with being explicit.

documentation/adr/0012-matlab-python-wrapper.md
```diff
@@ -0,0 +1,53 @@
...
+to expose Euphonic or Brille (or indeed any other Python program) to Matlab.
+The [proposed wrapper](https://github.com/mducle/horace-euphonic-interface/blob/light_wrapper/%2Beuphonic/light_python_wrapper.m)
+also handles data conversion and translation between Python and Matlab keyword arguments and some error checking.
+Finally, the wrapper also exposes the `pydoc` documentation system
+allowing users in Matlab to access the Python documentation transparently,
+and to use tab-completion to access the properties and methods of the Python class or object.
+
+To make it flexible (to allow for example the Euphonic Matlab interface to be more easily distributed on the File-Exchange),
+we propose to make this Python wrapper class a separate github repository which is either included as a submodule
+or purely in the build system (e.g. a cmake external project).
```
rebeccafair: Is this decision part of this ADR or a separate one? If it's another ADR it should be stated, and maybe give a placeholder link and link to the real ADR once its been written.

mducle: I'm not sure... I'd like to have a meeting between us and Greg, and then decide. I guess it should probably be a separate ADR.

documentation/adr/0012-matlab-python-wrapper.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Decision
+
+A decision will be taken at a forthcoming meeting.
+
+
+## Consequences
+
+The current [`horace-euphonic-interface`](https://github.com/pace-neutrons/horace-euphonic-interface/) and
+[`brillem`](https://github.com/brille/brillem/) code would need to be refactored to use the wrapper,
```
nickbattam-tessella: "would need to be refactored to use the wrapper" would benefit from some expansion. 

The linked PRs don't give any indication how major this refactor would be: is this a significant update to the structure and layout of the project or a few minor updates to APIs or addition of new parent classes?

mducle: Ah sorry... the refactoring is done in the PR, but this has not been accepted yet (so the "current" code is the pre-PR code). Becky, Greg and I are meeting later this week to discuss how to proceed with this.

mducle: The refactoring is "major" and amounts to removing all existing Matlab codes in `horace-euphonic-interface` and `brillem` and replacing them with Python code. This Python code would be wrapped so that it appears like a Matlab class to Matlab (methods and properties are transparently passed through with some automatic type conversions and syntax checking).

The reason to replace the Matlab code with Python is so that they work better with `pyHorace` - they would not need to be "compiled" (with `mcc`) so that is two less dependencies to be pulled in and compiled.

Originally I had wanted the replacement Python code put into `euphonic` and `brille` itself but these code are rather specific to neutron scattering use so it was decided that they did not belong in the parent projects.

documentation/adr/0012-matlab-python-wrapper.md
```diff
@@ -0,0 +1,65 @@
...
+We can force Matlab to launch the dependent Python interpreter `OutOfProcess`
+but this imposes a significant performance penalty
+(extensive testing was not done but Brille+SpinW runs about 10x slower than with `InProcess`). 
+
+
+## Proposal
+
+So, we propose that the Matlab interface to Euphonic and Brille would be through a Python wrapper class.
+There would be two implementation of this class.
+One (used by Matlab users) would call to `py.*` internally, whilst the other
```
cmarooney-stfc: This is a little ambiguous, although the context makes things cleaer. All Horace users are matlab users as Horace is written in matlab. You really mean users who invoke Horace natively from the Matlab command line. Not sure what the best way of saying that is, but I think it needs to be clearer.

mducle: Replaced it with "user interface (UI)" throughout.

rebeccafair (approved):
Looks good to me

PR #24, opened 2021-01-03
Add detailed description of pyHorace implementation

mducle:
A design document describing the prototype compiled Matlab for Python implementation of PACE.

### DIFF
--- a/python_interface/design/03_pyHorace_description.md
+++ b/python_interface/design/03_pyHorace_description.md
@@ -1,0 +1,269 @@
+# A description of the prototype pyHorace implementation
+**2020-12-29**
+
+
+## Table of Contents
+
+[Background](#background)
+
+[Compiled Matlab](#compiled-matlab)
+
+[Gateway functions and proxy classes](#gateway-functions-and-proxy-classes)
+
+[Calling Python from compiled Matlab](#calling-python-from-compiled-matlab)
+
+[Limitations and future work](#limitations-and-future-work)
+
+
+# Background
+
+PACE is a collection of several different programs written in both Matlab and Python (or C/C++)
+and meant to be used from both Matlab *and* Python in an *integrated* fashion.
+
+That is, users should be able to call a PACE program written in Python from a PACE program written in Matlab and vice versa,
+using a Matlab *or* Python interpreter.
+It is relatively straightforward for Matlab users to access the Python components of PACE using 
+[Matlab's built-in `py.*`](https://uk.mathworks.com/help/matlab/call-python-libraries.html) interface/namespace.
+This document, however, focuses on the case of Python users accessing the Matlab components of PACE.
+
+As detailed in an [architectural decision record](../../0010-use-compiled-matlab-for-python.md)
+we chose to use the Matlab Compiler toolbox to "compile" the Matlab parts of PACE
+which allows our users to use PACE as a Python module 
+(tentatively called `pyHorace`, but perhaps it should be `pyPace` or some such) 
+without needing a Matlab license.
+Instead they need to download and install a Matlab Components Runtime (MCR), a free but restricted distribution of Matlab,
+which can be used to run the "compiled" Matlab code only (and not normal m-files).
+
+This method, detailed in [two](01_pace_python_high_level_discussion.md) [previous](02_pace_python_implementation_discussion.md) design documents,
+is subjected to some restrictions imposed by the way the MCR and the `py.*` Python-bridge is implemented by Mathworks.
+Firstly, the `py.*` implementation _always_ spawns its own dependent Python interpreter.
+When called within a normal Matlab interpreter this is fine.
+However, when the Matlab code has been compiled and is being run within a _Python_ interpreter,
+this second (Matlab-launched) Python interpreter causes the parent Python interpreter to crash.
+Thus, `pyHorace` cannot use any Matlab code which refers to `py.*`.
+
+Instead, an alternative method for Matlab code to call Python, via a `mex` file using the CPython API, has been implemented.
+Together with a new framework which wraps Python code in a Matlab object intended for use by Euphonic and Brille,
+this allows the use of both Matlab and Python PACE interfaces with the minimum of changes to existing Matlab or Python codes.
+
+This documents describes the separate parts of the [prototype](https://github.com/mducle/hugo)
+minimum viable implementation of the PACE Python interface,
+and also the prototype Python class wrapper intended to be used by the PACE Matlab interface.
+
+We first detail how Matlab classes are wrapped in a Python proxy class which exposes their properties and methods to Python.
+We then move to how communications between PACE programs written in different languages is facilitated.
+Finally we describe remaining issues with the approach taken and how the prototype may be improved for production.
+
+
+# Compiled Matlab
+
+As noted in [ADR #10](../../documentation/adr/0010-use-compiled-matlab-for-python.md) the decision was taken
+to use "compiled" Matlab code for the Python interface to Horace/Herbert rather than to rewrite them in Python/C++.
+This imposes certain constraints on an "pyHorace" implementation:
+
+1. Only user written "gateway" m-file *functions* maybe directly accessed by Python (or executed on the commandline) in a compiled Matlab code.
+2. Matlab is run as a loaded library within the Python process ("in-process" execution).
+3. The Matlab help system is not available to compiled Matlab code.
+4. The Matlab MCR requires the use of its own standard libraries which implies restrictions on which compilers/Python versions can be used.
+
+Point 1 means that Matlab classes and standard Matlab functions cannot be *directly* created or called from Python 
+(a specific gateway function which constructs the class must be used instead).
+However, any Matlab code (and indeed the entire standard Matlab built-in library of functions and classes) maybe in "included" in the compiled library.
+The restrictions is that this "included" code may only be accessed by the user-written m-file *function* (not class)
+and not by the user calling those functions/classes from Python.
+Given this major limitation, the approach taken by the "Hugo" prototype is to use a gateway function,
+`call_method` which takes the *name* of a Matlab function, class or class method and a set of arguments and uses
+`feval` to call those "included" functions/class/methods.
+This is described in more detail in the [next section](#gateway-functions-and-proxy-classes).
+
+Point 2 means that the native Matlab way of calling Python (using the `py.*` namespace) cannot be used
+but also means that a `mex` file can be used for this purpose.
+This is described in detail in the [section on Python user functions](#calling-python-from-compiled-matlab).
+
+Point 3 means that if we would like users to be able to use `help(sqw)` or similar
+we would need to generate the help text and wrappers for each function / class we export (not currently implemented).
+This and point 4 and other limitations are discussed in the [limitations section](#limitations-and-future-work)
+
+
+# Gateway functions and proxy classes
+
+["pyHorace"](https://github.com/mducle/hugo) is based on a framework (["pySpinW"](https://github.com/spinw/pyspinw)) written for SpinW by S. Ward.
+The core of this is a Matlab "gateway" function, [`call_method.m`](https://github.com/mducle/hugo/blob/master/src/call_method.m#L101)
+which takes the name of a Matlab function and its arguments and uses `feval` to evaluate it.
+When compiled using the Matlab Compiler SDK as a Python module, it maybe accessed in Python using `<module_name>.call_method('function_name', object, [<arg_list>])`.
+To make it easier to use, this core gateway function and the Matlab Python module (called `horace`) is wrapped in a Python class,
+[`pyHorace.Matlab`](https://github.com/mducle/hugo/blob/master/pyHorace/Matlab.py#L60) whose `__getattr__()` special method is overloaded to call `call_method`.
+That is, when the user does:
+
+```
+from pyHorace import Matlab
+Matlab.<func_name>(*args)
+```
+
+The `Matlab.<func_name>` part causes Python to call `Matlab.__getattr__('func_name')`. 
+This returns a Python function which wraps `horace.call_method('func_name', [], *args)`.
+Both these steps are chained by Python so that uses can do `Matlab.cut_sqw('sqw_file', [], [], [], [-1,1])` transparently.
+Because the wrapper to `call_method` is through the `__getattr__` overload of a class, users always have to use the "dot" method-access notation.
+I.e., `from pyHorace.Matlab import *; cut_sqw('sqw_file', [], [], [], [-1,1])` will not work because `pyHorace.Matlab` is not a Python module but a class.
+In addition `help(pyHorace.Matlab.cut_sqw)` will also not work because this actually refers to a (dynamically constructed) Python wrapper function.
+Thus if we want the above two functionalities (accessing Matlab "included" functions/classes and help)
+we will need to write explicit Python wrappers for *every* Matlab function we want to expose this way.
+
+The other major feature of `pyHorace` is a Python [`MatlabProxyObject`](https://github.com/mducle/hugo/blob/master/pyHorace/MatlabProxyObject.py) class
+which wraps a Matlab class so that it appears transparent to Python users.
+That is, Python users can call the methods and access the properties of any Matlab class in Python using the "dot" notation as if they are in Matlab.
+This works by overloading the special Python methods `__getattr__` and `__setattr__`
+to call the equivalent Matlab special functions `subsref` and `subassgn`
+to access the methods/properties of a class.
+In addition, the Python `__repr__`, `__str__`, `__dir__` and `__doc__` methods are also overloaded
+to provide a description of the object, its methods and help text (except that the Matlab help system is not accessible in a compiled application).
+
+This wrapper is needed because any new-style (defined by `classdef`) Matlab object returned by `call_method` and its wrappers
+is opaque to Python and cannot be handled by purely Python functions.
+Old-style Matlab objects are treated as plain `struct`s and are automatically converted to Python `dict`s,
+losing any connection with the methods defined in their `@` folders.
+To overcome this, `pyHorace` includes a [`thinwrapper`](https://github.com/mducle/hugo/blob/master/src/thinwrapper.m) class which wraps an old style class.
+Whenever an old-style class (or [struct with an old style member](https://github.com/mducle/hugo/blob/master/src/has_thin_members.m))
+is returned by a Matlab function to Python, [`call_method`](https://github.com/mducle/hugo/blob/master/src/call_method.m#L115)
+instead [wraps](https://github.com/mducle/hugo/blob/master/src/wrap_obj.m) it (recursively) by storing the actual old-style object in the Matlab base namespace
+and returning to Python a `thinwrapper` (new-style) object.
+The only property of this object is the string name of the actual (old-style) object in the base namespace.
+This is used by [Python side wrappers](https://github.com/mducle/hugo/blob/master/pyHorace/MatlabProxyObject.py#L79)
+to allow users to transparently access the (old-style) Matlab objects as if they were Python objects.
+
+To facilitate all this wrapping, and to make the user syntax more natural, there is an automatic data conversion class,
+[`DataTypes`](https://github.com/mducle/hugo/blob/master/pyHorace/DataTypes.py),
+which converts data from Python to Matlab (`encode`) and vice versa (`decode`).
+Two particular use cases of this automatic conversion is to convert a Python list of numbers 
+[directly](https://github.com/mducle/hugo/blob/master/pyHorace/DataTypes.py#L42) into a Matlab numerical array
+(Matlab would otherwise convert any Python list into a *cell* array which would give an error if the Matlab function was not expecting this);
+and to [wrap `numpy` arrays](https://github.com/mducle/hugo/blob/master/pyHorace/TypeWrappers.py)
+such that no data copy is involved in the Python-to-Matlab conversion. 
+
+The `DataTypes` class also handles transparent conversion of the `MatlabProxyObject` and allows Python functions to be passed to Matlab
+where they can be evaluated through a `mex` function as explained in the next section.
+
+
+# Calling Python from compiled Matlab
+
+As noted in the [restrictions section](#compiled-matlab), Python code can be called in the normal Matlab interpreter using the `py.*` namespace.
+For example, `py.numpy.random.rand(int32(3),int32(3))` will create a 3x3 numpy array of random numbers in Matlab.
+This method, however, *must* not be used in compiled Matlab code intended to be used as a Python module,
+because it causes a second Python interpreter to started within the same process as the Matlab process 
+(which is actually the same process as the parent *Python* process because we are loading Matlab "in-process"). 
+Having two Python interpreters in the same process invariably causes a memory error and hard crash.
+There is no restriction on including `py.*` code in a compiled Matlab application 
+because they can be safely handled if the Matlab application is called directly from the commandline and not loaded as a Python module.
+A possible work-around is create the second Python interpreter (launched by `py.*`) in a new process ("out-of-process" execution). 
+However testing shows that this imposes huge performance penalties (using Brille becomes an order of magnitude slower).
+Furthermore for `pyHorace` the second interpreter is not connected with the user's interpreter
+so cannot know about any fitting/modelling functions they have defined.
+
+For these reasons, `pyHorace` opted to use a `mex` [file](https://github.com/mducle/hugo/blob/master/src/call_python.cpp) which is
+called by Matlab instead of the Python function input. 
+This mex file is linked to the CPython API so it can access the parent Python interpreter which is loaded into the same process.
+(Thus this method will not work with an out-of-process Python interpreter).
+
+First when the `DataTypes` class encounters a Python callable (a function or class which can be called), instead of passing this to Matlab,
+it stores the function reference in a [module-global dictionary](https://github.com/mducle/hugo/blob/master/pyHorace/FunctionWrapper.py#L21),
+and then creates a Matlab [`pythonFunctionWrapper`](https://github.com/mducle/hugo/blob/master/src/pythonFunctionWrapper.m) object
+which stores the (string) key to the Python function.
+Whenever `call_method` encounters a `pythonFunctionWrapper` object,
+it [wraps](https://github.com/mducle/hugo/blob/master/src/check_wrapped_function.m) this in a call to the mex `call_python` function with the correct key.
+Thus it appears to the user that Matlab is calling a Python function transparently (but actually happens is that Matlab calls `call_python`).
+`call_python` then [retrieves](https://github.com/mducle/hugo/blob/master/src/call_python.cpp#L390) the Python function reference from the module-global `dict` 
+and [calls](https://github.com/mducle/hugo/blob/master/src/call_python.cpp#L413) that function with the arguments given to it.
+
+The majority of the code in `call_python` is for data conversion between Matlab and Python.
+This is because `mex` functions may only take and return *Matlab* data types (Python objects are not allowed).
+Similarly the CPython `PyObject_Call` function requires that inputs to the called functions are all of the `PyObject` type.
+Thus `call_python` converts Matlab types to their Python equivalent and vice versa as per this table:
+
+|      Matlab    |      Python         | Note |
+| -------------- | ------------------- | ---- |
+| numeric array  | `numpy.ndarray`     |      |
+| numeric array  | `list` of numbers   | p2m  |
+| `char` array   | `str`               |      |
+| `string`       | `str`               |      |
+| `string` array | `list` of `str`     |      |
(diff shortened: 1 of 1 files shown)


python_interface/design/03_pyHorace_description.md
```diff
@@ -0,0 +1,269 @@
...
+
+That is, users should be able to call a PACE program written in Python from a PACE program written in Matlab and vice versa,
+using a Matlab *or* Python interpreter.
+It is relatively straightforward for Matlab users to access the Python components of PACE using 
+[Matlab's built-in `py.*`](https://uk.mathworks.com/help/matlab/call-python-libraries.html) interface/namespace.
+This document, however, focuses on the case of Python users accessing the Matlab components of PACE.
+
+As detailed in an [architectural decision record](../../0010-use-compiled-matlab-for-python.md)
+we chose to use the Matlab Compiler toolbox to "compile" the Matlab parts of PACE
+which allows our users to use PACE as a Python module
```
cmarooney-stfc: Being grammatically picky (sorry), "which allows" (singular) presumably refers to "compile" or possibly "use". However a natural reading might scan allows as going with "parts" (albeit plural). To clarify, maybe just start a new sentence and change "which" to ". This".

python_interface/design/03_pyHorace_description.md
```diff
@@ -0,0 +1,269 @@
...
+Finally we describe remaining issues with the approach taken and how the prototype may be improved for production.
+
+
+# Compiled Matlab
+
+As noted in [ADR #10](../../documentation/adr/0010-use-compiled-matlab-for-python.md) the decision was taken
+to use "compiled" Matlab code for the Python interface to Horace/Herbert rather than to rewrite them in Python/C++.
+This imposes certain constraints on an "pyHorace" implementation:
+
+1. Only user written "gateway" m-file *functions* maybe directly accessed by Python (or executed on the commandline) in a compiled Matlab code.
```
cmarooney-stfc: Is this MATLAB called from Python or Python called from MATLAB? The wording is a bit ambuguous. Assuming this is so, rewrite as "Only user-written 'gateway' m-file functions may be called directly from Python if they are to be called from compiled MATLAB code." Equivalent changes if it is actually the other way round".

python_interface/design/03_pyHorace_description.md
```diff
@@ -0,0 +1,269 @@
...
+2. Matlab is run as a loaded library within the Python process ("in-process" execution).
+3. The Matlab help system is not available to compiled Matlab code.
+4. The Matlab MCR requires the use of its own standard libraries which implies restrictions on which compilers/Python versions can be used.
+
+Point 1 means that Matlab classes and standard Matlab functions cannot be *directly* created or called from Python 
+(a specific gateway function which constructs the class must be used instead).
+However, any Matlab code (and indeed the entire standard Matlab built-in library of functions and classes) maybe in "included" in the compiled library.
```
cmarooney-stfc: Please separate "maybe" into "may be" - the meaning is completely different. Also remove in before included".

python_interface/design/03_pyHorace_description.md
```diff
@@ -0,0 +1,269 @@
...
+
+The `pyHorace` implementation, on the other hand will rely on `call_python` with Matlab-side and Python-side wrappers.
+On the Python side, a [`WrappedPythonClass`](https://github.com/mducle/hugo/blob/master/pyHorace/FunctionWrapper.py#L27)
+is used to wrap any Python class such that its objects are stored in a module-global dictionary and only a string reference is passed to Matlab.
+Additional functions, `get_obj_prop` and `call_obj_method` as well as wrapped constructors are defined
+and stored in a module-global dictionary for access by `call_python`.
+On the Matlab side, a [`pyclasswrapper`](https://github.com/mducle/hugo/blob/master/src/pyclasswrapper.m) 
+is used to wrap the these constructors and property access functions so as to present the same interface to the Matlab code as the pure Matlab case.
+The point of these indirections is that only string references rather than objects are passed between Matlab and Python
+but the behaviour is similar to creating objects and accessing their properties and methods directly.
```
rebeccafair: This section is a little confusing, it says the above mechanism works for user Python functions and Euphonic and that a wrapper is needed for the SpinW-Brille case, but then the implementation example given is for the horace-euphonic-interface.

It might be worth explaining why this is needed for the euphonic-horace-interface too (and if this implementation example is going to be expanded for use with SpinW-Brille)

rebeccafair (changes requested):
Just one main comment, otherwise a good summary and I think I understand it a lot more now! (although still not entirely)

python_interface/design/03_pyHorace_description.md
mducle: Sorry I was eliding two different things here together.

One is the mechanism for Matlab code to call Python when the user is using Matlab (the "implementation" linked - which points to the library I want both `horace-euphonic-interface` and `brillem` to use).

The second is the mechanism for compiled Matlab code to call Python internally when the user is using Python. This is the one that applies specifically to the SpinW-Brille interface.

I clarify this.

python_interface/design/03_pyHorace_description.md
```diff
@@ -0,0 +1,269 @@
...
+This is described in detail in the [section on Python user functions](#calling-python-from-compiled-matlab).
+
+Point 3 means that if we would like users to be able to use `help(sqw)` or similar
+we would need to generate the help text and wrappers for each function / class we export (not currently implemented).
+This and point 4 and other limitations are discussed in the [limitations section](#limitations-and-future-work)
+
+
+# Gateway functions and proxy classes
+
+["pyHorace"](https://github.com/mducle/hugo) is based on a framework (["pySpinW"](https://github.com/spinw/pyspinw)) written for SpinW by S. Ward.
```
nickbattam-tessella, line 93: This link references "Hugo" (the prototype) but talks of "pyHorace". The naming is unclear and the first link is simply to "prototype"

python_interface/design/03_pyHorace_description.md
```diff
@@ -0,0 +1,269 @@
...
+```
+
+The `Matlab.<func_name>` part causes Python to call `Matlab.__getattr__('func_name')`. 
+This returns a Python function which wraps `horace.call_method('func_name', [], *args)`.
+Both these steps are chained by Python so that uses can do `Matlab.cut_sqw('sqw_file', [], [], [], [-1,1])` transparently.
+Because the wrapper to `call_method` is through the `__getattr__` overload of a class, users always have to use the "dot" method-access notation.
+I.e., `from pyHorace.Matlab import *; cut_sqw('sqw_file', [], [], [], [-1,1])` will not work because `pyHorace.Matlab` is not a Python module but a class.
+In addition `help(pyHorace.Matlab.cut_sqw)` will also not work because this actually refers to a (dynamically constructed) Python wrapper function.
+Thus if we want the above two functionalities (accessing Matlab "included" functions/classes and help)
+we will need to write explicit Python wrappers for *every* Matlab function we want to expose this way.
```
nickbattam-tessella: Does that extend to wrappers for MATLAB class methods?

python_interface/design/03_pyHorace_description.md
```diff
@@ -0,0 +1,269 @@
...
+That is, Python users can call the methods and access the properties of any Matlab class in Python using the "dot" notation as if they are in Matlab.
+This works by overloading the special Python methods `__getattr__` and `__setattr__`
+to call the equivalent Matlab special functions `subsref` and `subassgn`
+to access the methods/properties of a class.
+In addition, the Python `__repr__`, `__str__`, `__dir__` and `__doc__` methods are also overloaded
+to provide a description of the object, its methods and help text (except that the Matlab help system is not accessible in a compiled application).
+
+This wrapper is needed because any new-style (defined by `classdef`) Matlab object returned by `call_method` and its wrappers
+is opaque to Python and cannot be handled by purely Python functions.
+Old-style Matlab objects are treated as plain `struct`s and are automatically converted to Python `dict`s,
```
nickbattam-tessella, line 131: The ongoing Horace refactor should be replacing all of the old-style classes with equivalent new-style ones; however having support for old-style classes is probably not a bad addition.

python_interface/design/03_pyHorace_description.md
```diff
@@ -0,0 +1,269 @@
...
+Whenever `call_method` encounters a `pythonFunctionWrapper` object,
+it [wraps](https://github.com/mducle/hugo/blob/master/src/check_wrapped_function.m) this in a call to the mex `call_python` function with the correct key.
+Thus it appears to the user that Matlab is calling a Python function transparently (but actually happens is that Matlab calls `call_python`).
+`call_python` then [retrieves](https://github.com/mducle/hugo/blob/master/src/call_python.cpp#L390) the Python function reference from the module-global `dict` 
+and [calls](https://github.com/mducle/hugo/blob/master/src/call_python.cpp#L413) that function with the arguments given to it.
+
+The majority of the code in `call_python` is for data conversion between Matlab and Python.
+This is because `mex` functions may only take and return *Matlab* data types (Python objects are not allowed).
+Similarly the CPython `PyObject_Call` function requires that inputs to the called functions are all of the `PyObject` type.
+Thus `call_python` converts Matlab types to their Python equivalent and vice versa as per this table:
```
nickbattam-tessella, line 188: I'm unclear how new-style classes are handled here.

python_interface/design/03_pyHorace_description.md
```diff
@@ -0,0 +1,269 @@
...
+Alternatively the `Matlab` and `MatlabProxyObject` classes can be modified to automatically
+convert Python keyword arguments to `('name', value)` pairs to be passed to Matlab if the Matlab code is modified to accept them.
+
+Limitation 3 comes from a restriction of the compiled Matlab output itself:
+Matlab does not allow access to its online documentation system from compiled code.
+The only way to work around this is to save the help text for every Matlab function or class we want to expose to Python to a text file
+and then to either generate wrappers for each (such as that used to overcome limitations 1 and 2)
+or to modify the `Matlab` and `MatlabProxyObject` classes to look up these help text.
+
+Limitation 4 is because the MPI routines needs to launch additional instances of the Matlab interpreter which is not possible from within Python.
```
nickbattam-tessella: This is a serious limitation(!). 

The creation of and execution of jobs through the MPI framework is a key element of the Horace performance improvements -- without parallelizing the algorithms there is only limited scope for performance improvements over what we currently have.

python_interface/design/03_pyHorace_description.md
```diff
@@ -0,0 +1,269 @@
...
+
+5. Must compile Matlab code for specific Matlab release.
+6. Must use a specific C++ compiler for the `mex` files.
+
+Restriction 5 means for example that a package created with Matlab R2020a can only be used with the MCR 2020a version.
+(Note that users must download and install the MCR to run `pyHorace` even if they alread have a Matlab license and Matlab installed
+unless they also have the Matlab Compiler toolbox installed).
+Matlab also only support specific C++ compilers for `mex` files.
+Using an older compiler will generate files which will work on a new Matlab version but the reverse is generally not true.
+This means that if we want to support the widest range of operating systems and Matlab versions,
```
nickbattam-tessella, line 311: The Horace build pipelines are creating and testing mex files with multiple MATLAB/platform combination and making all resulting binaries available on the GitHub releases page.

python_interface/design/03_pyHorace_description.md
mducle: Yeah... we need to decide on name... "Hugo" is just my whimsiness... the actual syntax uses `pyhorace`. But I think it should be called `pypace` or `pace-python` or some such...

python_interface/design/03_pyHorace_description.md
mducle: Yes, I think so - at least if we want to fully exploit the `pydoc` system for online help. 

If not, then it depends on whether we want to extend the syntax of that method. If the Matlab syntax accepts "keyword" arguments - (string,value) pairs - then we won't need to wrap the methods. 

I'm hoping we can just define a kind of specification and then autogenerate the wrappers...

python_interface/design/03_pyHorace_description.md
mducle: Matlab classes are currently not handled... It should certainly be possible to handle them (wrapped in a `MatlabProxyObject` Python wrapper) but this needs to be implemented (mainly because I've not needed to do this for any of the current functionality).

python_interface/design/03_pyHorace_description.md
mducle: As I mention below this should be possible to overcome. It would need changes to the `init` methods of `ClusterMPI` and `ClusterHerbert` (`ClusterParPool` should be fine I think). Essentially, instead of the code to launch a worker Matlab processes with `ProcessBuilder`, we would instead launch the compiled Matlab executables instead.

It's just not implemented yet.

python_interface/design/03_pyHorace_description.md
mducle: Yeah... the only issue is that Matlab is a few releases behind the latest compilers (especially on Linux), so e.g. won't be able to use some C++17 or C++20 features until a few years time.

python_interface/design/03_pyHorace_description.md
```diff
@@ -0,0 +1,269 @@
...
+This and point 4 and other limitations are discussed in the [limitations section](#limitations-and-future-work)
+
+
+# Gateway functions and proxy classes
+
+["pyHorace"](https://github.com/mducle/hugo) is based on a framework (["pySpinW"](https://github.com/spinw/pyspinw)) written for SpinW by S. Ward.
+The core of this is a Matlab "gateway" function, [`call_method.m`](https://github.com/mducle/hugo/blob/master/src/call_method.m#L101)
+which takes the name of a Matlab function and its arguments and uses `feval` to evaluate it.
+When compiled using the Matlab Compiler SDK as a Python module, it maybe accessed in Python using `<module_name>.call_method('function_name', object, [<arg_list>])`.
+To make it easier to use, this core gateway function and the Matlab Python module (called `horace`) is wrapped in a Python class,
```
cmarooney-stfc: Is this the right way round? It looks like the Python-language module called Matlab from what follows but this line looks like a Matlab-language module called Python.

python_interface/design/03_pyHorace_description.md
```diff
@@ -0,0 +1,269 @@
...
+[`pyHorace.Matlab`](https://github.com/mducle/hugo/blob/master/pyHorace/Matlab.py#L60) whose `__getattr__()` special method is overloaded to call `call_method`.
+That is, when the user does:
+
+```
+from pyHorace import Matlab
+Matlab.<func_name>(*args)
```
cmarooney-stfc: Is <func_name> actually coded with the angle brackets or do they just indicate that func_name should be substituted?

python_interface/design/03_pyHorace_description.md
```diff
@@ -0,0 +1,269 @@
...
+```
+
+The `Matlab.<func_name>` part causes Python to call `Matlab.__getattr__('func_name')`. 
+This returns a Python function which wraps `horace.call_method('func_name', [], *args)`.
+Both these steps are chained by Python so that uses can do `Matlab.cut_sqw('sqw_file', [], [], [], [-1,1])` transparently.
```
cmarooney-stfc: Not clear what chained means here. Do you mean that the single call Matlab.cut_sqw(...) handles all the getattr and call_method calls internally. May be simpler to say so explicitly.

python_interface/design/03_pyHorace_description.md
```diff
@@ -0,0 +1,269 @@
...
+Because the wrapper to `call_method` is through the `__getattr__` overload of a class, users always have to use the "dot" method-access notation.
+I.e., `from pyHorace.Matlab import *; cut_sqw('sqw_file', [], [], [], [-1,1])` will not work because `pyHorace.Matlab` is not a Python module but a class.
```
cmarooney-stfc: Do you mean e.g.? cut_sqw is just an example.

python_interface/design/03_pyHorace_description.md
cmarooney-stfc, line 131: Paragraph break should be before this line, not before "This wrapper". That separates the new-style and old-style class texts.

python_interface/design/03_pyHorace_description.md
```diff
@@ -0,0 +1,269 @@
...
+to provide a description of the object, its methods and help text (except that the Matlab help system is not accessible in a compiled application).
+
+This wrapper is needed because any new-style (defined by `classdef`) Matlab object returned by `call_method` and its wrappers
+is opaque to Python and cannot be handled by purely Python functions.
+Old-style Matlab objects are treated as plain `struct`s and are automatically converted to Python `dict`s,
+losing any connection with the methods defined in their `@` folders.
+To overcome this, `pyHorace` includes a [`thinwrapper`](https://github.com/mducle/hugo/blob/master/src/thinwrapper.m) class which wraps an old style class.
+Whenever an old-style class (or [struct with an old style member](https://github.com/mducle/hugo/blob/master/src/has_thin_members.m))
+is returned by a Matlab function to Python, [`call_method`](https://github.com/mducle/hugo/blob/master/src/call_method.m#L115)
+instead [wraps](https://github.com/mducle/hugo/blob/master/src/wrap_obj.m) it (recursively) by storing the actual old-style object in the Matlab base namespace
```
cmarooney-stfc: Not sure what you mean by recursively. Do these calls go through several levels of invocation if needed. Or do you just mean the 2 levels described here, in which case "successively" may be appropriate.

python_interface/design/03_pyHorace_description.md
```diff
@@ -0,0 +1,269 @@
...
+and returning to Python a `thinwrapper` (new-style) object.
+The only property of this object is the string name of the actual (old-style) object in the base namespace.
+This is used by [Python side wrappers](https://github.com/mducle/hugo/blob/master/pyHorace/MatlabProxyObject.py#L79)
+to allow users to transparently access the (old-style) Matlab objects as if they were Python objects.
+
+To facilitate all this wrapping, and to make the user syntax more natural, there is an automatic data conversion class,
+[`DataTypes`](https://github.com/mducle/hugo/blob/master/pyHorace/DataTypes.py),
+which converts data from Python to Matlab (`encode`) and vice versa (`decode`).
+Two particular use cases of this automatic conversion is to convert a Python list of numbers
```
cmarooney-stfc: Two .. use cases of ... conversion ARE to convert

python_interface/design/03_pyHorace_description.md
```diff
@@ -0,0 +1,269 @@
...
+| `string`       | `str`               |      |
+| `string` array | `list` of `str`     |      |
+| `struct`       | `dict`              |      |
+| `struct` array | `list` of `dict`    |      |
+| `cell` array   | `list`              |      |
+| `cell` array   | `tuple`             | p2m  |
+
+"p2m" indicates that conversion only applies from Python to Matlab 
+(from a Python `list` of numbers to a Matlab numeric array, and from a python `tuple` to a Matlab `cell` array).
+Other conversions apply in both directions (e.g. Matlab cell arrays are always converted to a Python `list` and vice versa).
```
cmarooney-stfc: Is this consistent? cell_array always converted to list and v.v., but with p2m cell_array is converted to tuple.  One of these must be wrong.

python_interface/design/03_pyHorace_description.md
```diff
@@ -0,0 +1,269 @@
...
+The conversion from a Matlab array to a Python array involves no data copies - the raw memory pointed to by Matlab is wrapped in a numpy `array` object.
+This is possible because that memory is only used in the Python function to be executed and not more widely.
+The reverse conversion (from a `numpy` array to a Matlab array) involves a data copy
+because Matlab `mex` does not support a shared data wrapping over memory which is not allocated by Matlab.
```
cmarooney-stfc, line 212: This is a bit unclear because it does not say who allocates the memory. If the receiving memory was allocated by matlab, would that work?

python_interface/design/03_pyHorace_description.md
```diff
@@ -0,0 +1,269 @@
...
+
+Note that Matlab objects, enumerations and sparse data types are currently not converted and will give an "`Unrecognised input`" error.
+Likewise, Python objects or other data types not listed above returned by the Python functions will give an "`Unknown return type`" error.
+In principle both these objects can be wrapped using the mechanisms described above for Matlab objects (through a `MatlabProxyObject`)
```
cmarooney-stfc: There appear to be 6 input types described here, so "both" is not appropriate. Did you mean "both these conversion directions"?

python_interface/design/03_pyHorace_description.md
```diff
@@ -0,0 +1,269 @@
...
+
+There are currently some limitations due to the architecture chosen for `pyHorace` which can be alleviated with more work:
+
+1. The Python syntax for any Matlab function/methods must exactly follow the Matlab syntax.
+2. Matlab functions can only be called as methods of the `Matlab` class and not independently (e.g. `without the m.* prefix`)
+3. Matlab help text cannot be accessed from Python.
+4. MPI routines don't work.
+
+Limitations 1 and 2 come from the choice, outlined in the [first section](#compiled-matlab) to use a gateway function which calls `feval`.
+This can be overcome by writing additional Python wrapper for each Matlab function or class we want to expose to Python.
```
cmarooney-stfc: by writing ... ADD "an" ... additional

python_interface/design/03_pyHorace_description.md
cmarooney-stfc: routineS needS ... one of the S needs to be removed

python_interface/design/03_pyHorace_description.md
```diff
@@ -0,0 +1,269 @@
...
+
+Limitation 3 comes from a restriction of the compiled Matlab output itself:
+Matlab does not allow access to its online documentation system from compiled code.
+The only way to work around this is to save the help text for every Matlab function or class we want to expose to Python to a text file
+and then to either generate wrappers for each (such as that used to overcome limitations 1 and 2)
+or to modify the `Matlab` and `MatlabProxyObject` classes to look up these help text.
+
+Limitation 4 is because the MPI routines needs to launch additional instances of the Matlab interpreter which is not possible from within Python.
+The most straight-forward way to overcome this limitation is to compile Horace as a stand-alone executable, with the `worker_v2.m` function as the gateway function. 
+This executable can then be launched and will communicate with the main process using files or MPI as usual. (This has not been tested).
```
cmarooney-stfc: This sounds like a critical defect. Is the MPI in Horace or Euphonic/Brille/3rd party? For Horace, does this prevent the current optimization work using MPI?

python_interface/design/03_pyHorace_description.md
```diff
@@ -0,0 +1,269 @@
...
+
+In addition, however, there are also limitations imposed by Matlab which cannot be alleviated:
+
+5. Must compile Matlab code for specific Matlab release.
+6. Must use a specific C++ compiler for the `mex` files.
+
+Restriction 5 means for example that a package created with Matlab R2020a can only be used with the MCR 2020a version.
+(Note that users must download and install the MCR to run `pyHorace` even if they alread have a Matlab license and Matlab installed
```
cmarooney-stfc: alreadY

python_interface/design/03_pyHorace_description.md
```diff
@@ -0,0 +1,269 @@
...
+unless they also have the Matlab Compiler toolbox installed).
+Matlab also only support specific C++ compilers for `mex` files.
```
cmarooney-stfc: supportS

python_interface/design/03_pyHorace_description.md
```diff
@@ -0,0 +1,269 @@
...
+Using an older compiler will generate files which will work on a new Matlab version but the reverse is generally not true.
```
cmarooney-stfc, line 310: newER

python_interface/design/03_pyHorace_description.md
```diff
@@ -0,0 +1,269 @@
+# A description of the prototype pyHorace implementation
```
cmarooney-stfc, line 1: Overall comments.
1. Please see the comment on MPI at the end and say if it is as critical as I think it may be.
2. While the overall structure is good, I would really like to see a vanilla code example showing the major features going from pyHorace to matlab and back to (e.g. Euphonic) python. The details of why a particular strategy is used obscures clarity in seeing what the final product looks like.
Thanks for doing this -very helpful

cmarooney-stfc (changes requested):

python_interface/design/03_pyHorace_description.md
mducle: Ah sorry. This is Matlab called from Python. I'll make the changes.

python_interface/design/03_pyHorace_description.md
mducle: It should be "the Python module (called `horace`) which contains the compiled Matlab archive".

python_interface/design/03_pyHorace_description.md
mducle: It should be substituted for the actual function name.

python_interface/design/03_pyHorace_description.md
mducle: Yes. What I meant to say is that it's a multipart process which is chained together by both Python and the wrappers:

1. In the first part, `Matlab.cut_sqw` - Python converts this to `Matlab.__getattr__('cut_sqw')`.
2. We overload `Matlab.__getattr__(func_name)` to return a function reference (an anonymous function really) which wraps the call to `horace.call_method('cut_sqw', [], *args)` with some type conversions on the inputs.
3. The next part `(*args)` of `Matlab.cut_sqw(*args)` is then interpreted by Python as calling the referenced function (the anonymous function which wraps  `call_method`) with the given arguments.

But you're right that just saying it's all handled internally is easier.

python_interface/design/03_pyHorace_description.md
mducle: Yes, you're right that it should be "successively", since old-style members of an old-style class don't need to be wrapped, just old-style fields of a struct (this was the issue with `multifit`).

python_interface/design/03_pyHorace_description.md
mducle: So going from Matlab to Python I chose to always convert a cell array to a `list`.

Going from Python to Matlab both `tuple` and `list` are converted to cell arrays.

E.g. this converter will never *create* a Python `tuple` from any Matlab types but accepts `tuple`s as inputs. 

This is because Matlab doesn't have any equivalent to a tuple.

I'll redo the table to make it clearer.

python_interface/design/03_pyHorace_description.md
mducle: Yes.

So there are two factory functions defined by the `mex` API: `CreateArray` and `CreateArrayFromBuffer`. `CreateArray` takes raw memory pointers or iterators and copies the data returning a Matlab `Array` object. `CreateArrayFromBuffer` in principle takes a smart pointer (`std::unique_ptr`) and steals the reference which means it doesn't do a copy. 

However, [Matlab only supports](https://uk.mathworks.com/matlabcentral/answers/514456-using-a-custom-deleter-in-a-buffer_ptr_t-in-arrayfactory-createarrayfrombuffer) the use case where you allocate the memory pointer by the `std::unique_ptr` using its own `CreateBuffer` function. So, if you allocate the buffer yourself and then use `CreateArrayFromBuffer` the code will compile and run but then will result in a segfault when Matlab tries to delete the memory.

This is because although everything is run within a single process and shares the same memory, the base Matlab interpreter, the mex file and the Python interpreters all have different heaps. In principle, `std::unique_ptr` allows you to overcome this by passing the pointer to the correct heap deallocator along with the pointer. However, for its own reasons Matlab does not honour this and overwrites the passed deallocator with its own function (which can only deallocate memory allocated in its own heap). To be fair it does `typedef` the `std::unique_ptr` to its own `buffer_ptr` type so you shouldn't expect it to obey the standard behaviour but it's still annoying...

python_interface/design/03_pyHorace_description.md
mducle: Yes.

python_interface/design/03_pyHorace_description.md
mducle: As I mentioned I haven't tested it, and for some reason the current implementation is not working for me on Matlab (on Linux) anyway...

So the first thing we need to do is to create the "compiled" workers. I see that there is some stub code at the moment so this needs to be implemented. I might have time next month for this if Alex is too busy.

Once this is done, I think that in principle it should work with the Python interface as is, or with minimum changes. For the file based stuff we might have to be careful about paths - it seems to use some Matlab settings folder which might not exist or be writable in the compiled Matlab / Python interface. 

For the MPI backend itself... it's rather system dependent unfortunately and the issue is that once we've compiled the Matlab code we can't change it for different systems. I'm not exactly clear on how it's implemented in Herbert but it might need to be refactored to not use `mex` but to load a library (`.so` or `.dll` C-style) which can be swapped out (the Matlab "compiler" includes `mex` files in the compiled archive so they cannot be changed). This way we could accomodate the different MPI implementations without providing a horde of different `mex` files in the distribution... I think I need to talk to Alex about this...

PR #25, opened 2021-01-03
Add description of options for pyHorace distribution

mducle:
A design document outlining the options for packaging `pyHorace` for distribution to users.

### DIFF
--- a/python_interface/design/04_pyHorace_packaging.md
+++ b/python_interface/design/04_pyHorace_packaging.md
@@ -29,6 +29,11 @@ This is because PyPI has a limit of 100MB per package and Anaconda has a limit o
 An alternative to bundling the MCR is to include a step to automatically download and install an MCR during installation or on first use.
 Doing this during installation limits the options to 1 or possibly 3 
 (PyPI/`pip` does not allow to execute a custom script during installation, whilst it _may_ be possible with `conda`).
+Alternatively, the Python package can provide a installation commandline tool.
+On import, if `pyHorace` detects that the MCR is not install it will fail with a `RuntimeError`
+which will tell the user to run the commandline tool to install the MCR,
+or to provide the path to where they installed it (if non-standard location)
+ in the `LD_LIBRARY_PATH` (Linux) or `PATH` (Windows) environment variable.
 
 Finally, because `pyHorace` is Python package it requires a version of Python to be installed.
 However, there are many possible Python distributions which users can install, depending on their OS.
--- a/python_interface/design/04_pyHorace_packaging.md
+++ b/python_interface/design/04_pyHorace_packaging.md
@@ -1,0 +1,61 @@
+# A description of the prototype pyHorace implementation
+**2021-1-3**
+
+
+# Background
+
+`pyHorace` will be a Python module which bundles together PACE programs written in Matlab for use by Python users without a Matlab license.
+
+This document describes the options for packaging it for distribution to users.
+
+
+# pyHorace packaging options
+
+There are a few ways to distribute `pyHorace` which we need to decide on:
+
+1. OS-specific packaging system
+2. The Python Package Index (PyPI)
+3. The Anaconda (`conda`) system
+
+Option 1 would be to create packages for specific OS (e.g. `msi` for Windows, `dmg` for Mac OS, `deb` for Debian/Ubuntu, `rpm` for Red Hat/CentOS etc.).
+These packages would be installed by users as they would a non-Python program.
+Options 2 and 3 are specifically for Python packages (or packages meant to be used with Python).
+
+A further decision is whether we should "bundle" the MCR with the `pyHorace` distribution or not.
+The MCR is a very large archive (~2.8GB for R2020b) because it basically contains the full Matlab distribution.
+If we want to distribute the MCR with the package, we are restricted to option 1 and self-hosting.
+This is because PyPI has a limit of 100MB per package and Anaconda has a limit of 1GB per package.
+
+An alternative to bundling the MCR is to include a step to automatically download and install an MCR during installation or on first use.
+Doing this during installation limits the options to 1 or possibly 3 
+(PyPI/`pip` does not allow to execute a custom script during installation, whilst it _may_ be possible with `conda`).
+
+Finally, because `pyHorace` is Python package it requires a version of Python to be installed.
+However, there are many possible Python distributions which users can install, depending on their OS.
+The advantage of option 2 and 3 over 1 is that they are tied to a specific Python installation,
+whilst in option 1 we need to determine which Python to install to.
+This is usually only a problem on Windows because on Linux there is usually a pre-installed system Python installation which can be used.
+On Windows, on the other hand, users will usually not have Python installed or may have multiple different (conflicting) Python installations.
+(This is the reason why Mantid bundles a version of Python with its Windows package).
+
+### Option 1 (own packaging) pros and cons
+
+- Most flexibility, easiest user experience. 
+- Most developer effort needed (initial development and ongoing maintenance).
+- Needs hosting infrastructure.
+- May need to bundle own Python installation on Windows.
+
+### Option 2 (PyPI) pros and cons
+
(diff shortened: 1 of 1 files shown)


python_interface/design/04_pyHorace_packaging.md
```diff
@@ -0,0 +1,61 @@
...
+
+1. OS-specific packaging system
+2. The Python Package Index (PyPI)
+3. The Anaconda (`conda`) system
+
+Option 1 would be to create packages for specific OS (e.g. `msi` for Windows, `dmg` for Mac OS, `deb` for Debian/Ubuntu, `rpm` for Red Hat/CentOS etc.).
+These packages would be installed by users as they would a non-Python program.
+Options 2 and 3 are specifically for Python packages (or packages meant to be used with Python).
+
+A further decision is whether we should "bundle" the MCR with the `pyHorace` distribution or not.
```
nickbattam-tessella, line 24: Is distribution of the MCR allowed under the terms of the MATLAB licences? When I last looked into this (in around 2012) we concluded that the end user would need to download and install the MCR themselves.

mducle: This is not entirely clear. If you use the `deploytool` to create a distribution, there is an option to include the MCR with it (making a huge file). The [`MATLAB RUNTIME LICENSE`](https://raw.githubusercontent.com/teknowledgist/Chocolatey-packages/master/manual/mcr-r2019a/MCR_license.txt) itself doesn't say one way or the other whether it is allowed or not (it doesn't explicitly allow or forbid it). I think that given the use in the `deploytool` and the principle that if it is not expressly forbidden, then it is permitted then we can bundle it... but IANAL.

Personally I think that we should *not* bundle the MCR (it will result in huge package artefacts and vastly increase our storage needs).

rebeccafair (commented):
Looks good to me! There may be some more points to add based on ease of testing/versioning but I don't know enough about how CI would work for each of the options to say which would make this easier or more difficult at this point.

python_interface/design/04_pyHorace_packaging.md
nickbattam-tessella: I tend to agree -- unless there is a compelling need to it will be much simpler to inform end-users how to get a compatible version. The only issue that may arise is if we, for some reason, are locked to an old version of the MCR which becomes hard to obtain. I would hope that won't happen and is a problem for another day.

mducle: Mathworks [provides runtime back to 2012a](https://uk.mathworks.com/products/compiler/matlab-runtime.html) so I think it should be ok. I would suggest that we stick to a single MCR version for longer than the 6 months Matlab release cycle, since we often don't need the new features, and it would mean that users would not have to update the MCR as often. For now I would suggest to stick to R2019b as this is setup on our CI servers and has all the features we need (especially support for C-style row major memory layouts), and to only update the version every 2-3 years (perhaps with every major version change of Horace).

python_interface/design/04_pyHorace_packaging.md
```diff
@@ -0,0 +1,61 @@
...
+
+### Option 2 (PyPI) pros and cons
+
+- Standard way to distribute Python packages.
+- Relatively easy / little developer effort to set up. 
+- Self-hosting not needed / minimal maintenance needed.
+- Users need to download and install Matlab MCR themselves.
+- Users need to have installed Python already themselves (mainly issue on Windows).
+
+### Option 3 (Anaconda) pros and cons
```
cmarooney-stfc, line 63: This seems written to support a preference of 3>2>1. No problem with this, just a couple of comments.
- from previous text, Anaconda supports a large package size, would we need to reserve this in case we get larger?
- about 2 years ago I was unable to install Ananconda properly as the per-user version had one set of permission problems, and the all-user version the other set. This went away the next time Anaconda released. Might suggest that tying us to Ananconda could lead to problems every so often?

python_interface/design/04_pyHorace_packaging.md
```diff
@@ -0,0 +1,61 @@
...
+
+- Widely used by scientific Python users (and probably most popular way to install Python on Windows).
```
nickbattam-tessella, line 65: If we release through Anaconda, is access easy for a PyPI based user (specifically on Linux)?

mducle: Unfortunately not. 

But the reverse is true (Anaconda can easily access PyPI but users using PyPI would need to install Anaconda separately). This is because Anaconda is a self contained packaging system which just happens to include Python amongst other things whilst PyPI is only for Python packages (so Anaconda packages are a superset of Python packages).

python_interface/design/04_pyHorace_packaging.md
mducle: Yeah... it might lead to some problems. I haven't had that problem myself (I tend to just install for my own user).

python_interface/design/04_pyHorace_packaging.md
mducle: Long time now... but this can now be resolved by using the [`compiler.package.installer`](https://uk.mathworks.com/help/compiler/compiler.package.installer.html) command to create a installable package which will download only the components of the MCR needed directly from Mathworks. This saves us bundling the MCR and also presents users with the "Do you accept this license" screen.

python_interface/design/04_pyHorace_packaging.md
```diff
@@ -0,0 +1,61 @@
...
+Options 2 and 3 are specifically for Python packages (or packages meant to be used with Python).
+
+A further decision is whether we should "bundle" the MCR with the `pyHorace` distribution or not.
+The MCR is a very large archive (~2.8GB for R2020b) because it basically contains the full Matlab distribution.
+If we want to distribute the MCR with the package, we are restricted to option 1 and self-hosting.
+This is because PyPI has a limit of 100MB per package and Anaconda has a limit of 1GB per package.
+
+An alternative to bundling the MCR is to include a step to automatically download and install an MCR during installation or on first use.
+Doing this during installation limits the options to 1 or possibly 3 
+(PyPI/`pip` does not allow to execute a custom script during installation, whilst it _may_ be possible with `conda`).
```
mducle, line 33:
```suggestion
(PyPI/`pip` does not allow to execute a custom script during installation, whilst it _may_ be possible with `conda`).
Alternatively, the Python package can provide a installation commandline tool.
On import, if `pyHorace` detects that the MCR is not install it will fail with a `RuntimeError`
which will tell the user to run the commandline tool to install the MCR,
or to provide the path to where they installed it (if non-standard location)
 in the `LD_LIBRARY_PATH` (Linux) or `PATH` (Windows) environment variable.
```

python_interface/design/04_pyHorace_packaging.md
```diff
@@ -0,0 +1,61 @@
...
+
+Option 1 would be to create packages for specific OS (e.g. `msi` for Windows, `dmg` for Mac OS, `deb` for Debian/Ubuntu, `rpm` for Red Hat/CentOS etc.).
+These packages would be installed by users as they would a non-Python program.
+Options 2 and 3 are specifically for Python packages (or packages meant to be used with Python).
+
+A further decision is whether we should "bundle" the MCR with the `pyHorace` distribution or not.
+The MCR is a very large archive (~2.8GB for R2020b) because it basically contains the full Matlab distribution.
+If we want to distribute the MCR with the package, we are restricted to option 1 and self-hosting.
+This is because PyPI has a limit of 100MB per package and Anaconda has a limit of 1GB per package.
+
```
mducle, line 28:
```suggestion

Recently (since R2020a), Matlab introduced a command, [`compiler.package.installer`](https://uk.mathworks.com/help/compiler/compiler.package.installer.html) which can create an installer executable which will only download the components / toolboxes needed by the application. For `pyHorace` this reduces the download size to ~500MB, since we only need the core computation and Python components. It is not possible to bundle this reduced sized MCR with our application however - it seems the only way is to create an installer using `compiler.package.installer`.

```

